PR #2, opened 2021-02-03
Feature/devsu 1427 incorporate templates in rendering reports

Anna-log7:
Re-opening PR after moving from bitbucket

This will allow for custom templates to be used to render reports including sidebar sections, print sections, and header images


Anna-log7:
Note @kttkjl I fixed the PR comments from bitbucket

kttkjl (approved):

PR #3, opened 2021-02-03
Feature/window variables

creisle:
Improvements
- Use dynamic post-build variables
- Set docker publish workflow to publish to ipr client dockerhub repo


app/components/NavBar/index.js
```diff
@@ -41,7 +41,7 @@ class NavBar {
       bindToController: true,
       controllerAs: 'vm',
       locals: {
-        CONFIG,
+        ENV: CONFIG.MISC.ENV,
```
Anna-log7: This should pass in the `CONTACT_EMAIL` and `CONTACT_TICKET_URL` instead of the `ENV`.
`$mdDialog` controllers can't access `window` properly so those values in the feedback component won't render.

I can make this change if you'd like since it's angular

Anna-log7: Fixed

Anna-log7 (approved):

PR #10, opened 2021-02-09
Feature/devsu 1403 refactor germline reports page

Anna-log7:
For use with the API branch here: https://github.com/bcgsc/pori_ipr_api/pull/8

- Changed germline boards page to use React instead of AngularJS
- Added custom filter to the exported header


app/views/GermlineView/components/Board/columnDefs.js
```diff
@@ -0,0 +1,69 @@
+import { formatDate } from '../../../../utils/date';
```
kttkjl, line 1: could probably update the path here to the new @

app/views/GermlineView/components/Board/components/ApiPaginatedTable/components/CheckboxCell/index.scss
```diff
@@ -0,0 +1,3 @@
+.checkbox-cell {
+  padding: 0 !important;
```
kttkjl, line 2: are there absolutely no other way than to use !important?

app/views/GermlineView/components/Board/columnDefs.js
kttkjl: oh wait this is waiting on the other PR to be merged in ... guess we can just fix it as we go

Anna-log7: The other PR is just so it doesn't show as a linting error - it still works. I can add it here

app/views/GermlineView/components/Board/components/ApiPaginatedTable/components/CheckboxCell/index.scss
Anna-log7: It didn't like it without in testing, but now it seems fine ?_?
Anyways, I removed it

app/views/GermlineView/components/Board/index.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { useEffect, useState } from 'react';
+
+import ApiPaginatedTable from './components/ApiPaginatedTable';
+import api from '../../../../services/api';
+import columnDefs from './columnDefs';
+import ParamsContext from './components/ParamsContext';
+
+const Board = (): JSX.Element => {
+  const [reports, setReports] = useState([]);
```
kttkjl, line 9: making a note here that this is typed any, but reports have a way complicated structure right?

kttkjl (approved):

app/views/GermlineView/components/Board/index.tsx
Anna-log7: It's not too complicated for germline reports so I'll add it in

kttkjl (approved):

PR #14, opened 2021-02-11
Add demo mode descriptions to sections

creisle:
Improvements
- Add demo-only information alerts to some details sections

### DIFF
--- a/app/views/ReportView/components/MutationSignatures/index.jsx
+++ b/app/views/ReportView/components/MutationSignatures/index.jsx
@@ -5,6 +5,7 @@ import orderBy from 'lodash.orderby';
 import { LinearProgress, Typography } from '@material-ui/core';
 import { useSnackbar } from 'notistack';
 
+import DemoDescription from '@/components/DemoDescription';
 import DataTable from '@/components/DataTable';
 import ReportContext from '../../../../components/ReportContext';
 import EditContext from '@/components/EditContext';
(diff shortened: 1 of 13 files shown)


config/webpack/webpack.dev.js
```diff
@@ -25,7 +25,7 @@ const devConfig = {
         CONTACT_EMAIL: 'ipr@bcgsc.ca',
         CONTACT_TICKET_URL: 'https://www.bcgsc.ca/jira/secure/CreateIssue!default.jspa',
         PUBLIC_PATH: '/',
-        IS_DEMO: true,
+        IS_DEMO: false,
```
creisle, line 28: this was accidentally committed as true originally. Should be false so that write operations are not disabled on iprdev. To test this you'll want to manually set it true

Anna-log7: Ha I was actually just about to message you about this after running into it myself

app/views/ReportView/components/MutationSignatures/index.jsx
```diff
@@ -161,8 +169,8 @@ const MutationSignatures = () => {
           />
         </>
       ) : (
-        <LinearProgress />
-      )}
+          <LinearProgress />
+        )}
```
Anna-log7, line 173: Indentation was added here resulting in a linting error

Anna-log7 (changes requested):

app/views/ReportView/components/MutationSignatures/index.jsx
creisle: weird, that was auto-fixed by the linter... do we have different settings somehow? will compare offline

creisle: note: we determined that something odd is going on where running eslint via command line doesn't show this error but sometimes it does in vscode. @Anna-log7  is currently strengthening the linting rules in another branch which should hopefully resolve this

Anna-log7 (approved):

PR #25, opened 2021-02-19
Bugfix/devsu 1450 fix msi scatter

Anna-log7:
This is to fix the sizing of the 1000x1000 msi scatter image. I added props to the `Image` component to allow it to be set to a custom size which fixes this issue.

### DIFF
--- a/app/components/AuthenticatedRoute/index.jsx
+++ b/app/components/AuthenticatedRoute/index.jsx
@@ -48,14 +48,15 @@ const AuthenticatedRoute = ({
     <>
       <Route
         {...rest}
-        render={props => (<ChildComponent admin={admin} {...props} />)}
+        render={(props) => (<ChildComponent admin={admin} {...props} />)}
       />
     </>
   );
 };
 
 AuthenticatedRoute.propTypes = {
-  component: PropTypes.node.isRequired,
+  // eslint-disable-next-line react/forbid-prop-types
+  component: PropTypes.object.isRequired,
   // eslint-disable-next-line react/forbid-prop-types
   location: PropTypes.object,
   admin: PropTypes.bool,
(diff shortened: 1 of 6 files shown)


app/components/AuthenticatedRoute/index.jsx
```diff
@@ -48,14 +48,15 @@ const AuthenticatedRoute = ({
...
+        render={(props) => (<ChildComponent admin={admin} {...props} />)}
       />
     </>
   );
 };
 
 AuthenticatedRoute.propTypes = {
-  component: PropTypes.node.isRequired,
+  // eslint-disable-next-line react/forbid-prop-types
+  component: PropTypes.object.isRequired,
```
kttkjl, line 59: would `PropTypes.element` work better?

app/components/AuthenticatedRoute/index.jsx
```diff
@@ -48,14 +48,15 @@ const AuthenticatedRoute = ({
...
   // eslint-disable-next-line react/forbid-prop-types
   location: PropTypes.object,
```
kttkjl, line 61: I thought we had defined a location propType before

Anna-log7: Not that I see, no

app/components/AuthenticatedRoute/index.jsx
Anna-log7: It complains with element too

app/components/AuthenticatedRoute/index.jsx
kttkjl: Ohh, graphKB then ... out of context for this PR, but 

```
const LocationPropType = PropTypes.shape({
  hash: PropTypes.string,
  key: PropTypes.string,
  pathname: PropTypes.string,
  search: PropTypes.string,
});
```
is what's defined in graphKB

kttkjl (approved):

PR #31, opened 2021-03-01
Feature/devsu 1077 add error snackbar by converting to react

Anna-log7:
The purpose of this PR is to get the error snackbar working in the `GeneViewer` component. To do this, the component needs access to the react context.
The `DataTable`s that were rendered by Angular wouldn't be able to access this context, thus the snackbar wouldn't appear. In this PR I've converted the somatic sections to full react. These were the last sections to have an angular `DataTable`, so I was able to remove the angular specific code from there including the `reportIdent` prop, the `theme` prop/`ThemeProvider`


kttkjl (changes requested):
Just the type comment and merge conflicts

kttkjl (approved):

PR #36, opened 2021-03-12
Update query parameters when getting reports

brayden-bcgsc:


app/views/ReportsView/index.jsx
```diff
@@ -30,13 +30,10 @@ function ReportsTableComponent(props) {
...
-    const opts = {
-      all: true,
-      states: 'ready,active,uploaded,signedoff,archived,reviewed',
-    };
+    const opts = {};
 
-    if (adminUser) {
-      opts.states = 'ready,active,uploaded,signedoff,archived,reviewed,nonproduction';
+    if (!adminUser) {
+      opts.states = 'ready,active,uploaded,signedoff,archived,reviewed';
```
Anna-log7, line 36: If no states are specified does it return them all?

brayden-bcgsc: Not specifying any states returns all reports regardless of state.

Anna-log7 (approved):

PR #43, opened 2021-03-30
Bugfix/devsu 1473 fix service worker registration

Anna-log7:
The service worker was never properly being registered on prod. This PR fixes that by using the workbox-webpack-plugin on prod.

Unfortunately since this is prod only it will be hard to test. You could copy the prod config part into the common webpack file and that should work

### DIFF
--- a/config/webpack/webpack.prod.js
+++ b/config/webpack/webpack.prod.js
@@ -3,6 +3,7 @@ const webpack = require('webpack');
 const merge = require('webpack-merge');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 const path = require('path');
+const { GenerateSW } = require('workbox-webpack-plugin');
 const common = require('./webpack.config.js');
 
 const prodConfig = {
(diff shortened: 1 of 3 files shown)


kttkjl:
> The service worker was never properly being registered on prod. This PR fixes that by using the workbox-webpack-plugin on prod.
> 
> Unfortunately since this is prod only it will be hard to test. You could copy the prod config part into the common webpack file and that should work

does running npm run start:prod not work?

Anna-log7:
> 
> 
> > The service worker was never properly being registered on prod. This PR fixes that by using the workbox-webpack-plugin on prod.
> > Unfortunately since this is prod only it will be hard to test. You could copy the prod config part into the common webpack file and that should work
> 
> does running npm run start:prod not work?
> 
> actually I just swapped the keycloak urls around and that seemed to have the service worker running under webpack-dev-server
> 
> though I'm not seeing a service worker registered when running locally -- according to google I'll need to have this run under https?
> 
> testing on an SSL-certified domain atm ... hold on

Yes it'll have to be run as HTTPS in order for the service worker library to be loaded which is why local won't work

kttkjl:
Service worker looks registered

kttkjl (approved):

PR #44, opened 2021-03-31
Feature/devsu 1431 add msi data to summary

Anna-log7:
This is to handle the MSI data in the genomic summary that will be loaded into IPR soon. I also added full typing to the genomic summary


app/views/ReportView/components/GenomicSummary/index.tsx
```diff
@@ -13,21 +12,26 @@ import { useSnackbar } from 'notistack';
...
+import {
+  MsiType,
+  PatientInformationType,
+  GeneVariantType,
+  TumourSummaryType,
+  MicrobialType,
+} from './types.d';
+import { MutationBurdenType, ComparatorType } from '../MutationBurden/types.d';
+import MutationSignatureType from '../MutationSignatures/types.d';
+import { ImmuneType } from '../Immune/types';
```
kttkjl, line 34: weird how this works, but `.d` is required of all the other types ...

app/components/Image/index.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { useState, useCallback } from 'react';
 import { Typography, Fade } from '@material-ui/core';
 
-import ImageType from './types';
+import ImageType from './types.d';
```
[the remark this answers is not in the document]
Anna-log7: oh that probably would. It was working on/off so I never thought about it really

kttkjl (commented):
- there are inconsistencies with imports on `.d` files, some of them work, some of them don't

- MSI addition works

kttkjl (changes requested):
there are inconsistencies with imports on .d files, some of them work without the .d in the path, some of them don't

MSI addition works

Github reviews are still weird to ...

app/views/ReportView/components/GenomicSummary/index.tsx
Anna-log7: I've removed the `.d` extension now

kttkjl (approved):

kttkjl (approved):

PR #49, opened 2021-04-26
Feature/devsu 1496 refactor settings to use react

Anna-log7:
- Changed the settings section from angular to react


app/views/ReportView/components/Settings/components/AddUserDialog/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+import UserAutocomplete from '@/components/UserAutocomplete';
+import ReportContext from '@/components/ReportContext';
+import snackbar from '@/services/SnackbarUtils';
+import { UserType } from '@/common';
+
+import './index.scss';
+
+type AddUserDialogProps = {
+  isOpen: boolean;
+  onAdd: (user?: UserType) => void;
```
kttkjl, line 26: is user unused? `onAdd` doesn't seem to be calling with a user in any of these instances

kttkjl (commented):
Code looks fine, just merge conflicts

app/views/ReportView/components/Settings/components/AddUserDialog/index.tsx
Anna-log7: I've removed the param

kttkjl (approved):

PR #58, opened 2021-05-20
Handle an array of images from the api

Anna-log7:
For use with: https://github.com/bcgsc/pori_ipr_api/pull/54
This is so images are processed as arrays rather than objects

### DIFF
--- a/app/views/ReportView/components/CopyNumber/index.tsx
+++ b/app/views/ReportView/components/CopyNumber/index.tsx
@@ -29,7 +29,7 @@ const CopyNumber = (): JSX.Element => {
   const { report } = useContext(ReportContext);
 
   const [isLoading, setIsLoading] = useState(true);
-  const [images, setImages] = useState<Record<string, ImageType>>();
+  const [images, setImages] = useState<ImageType[]>([]);
   const [cnvs, setCnvs] = useState<CopyNumberType[]>([]);
   const [groupedCnvs, setGroupedCnvs] = useState({
     clinical: [],
@@ -128,14 +128,12 @@ const CopyNumber = (): JSX.Element => {
       {!isLoading ? (
         <>
           <Typography variant="h3" className="copy-number__title">Summary of Copy Number Events</Typography>
-          {images?.['cnvLoh.circos'] && (
-            <div className="copy-number__circos">
-              <Image
-                image={images['cnvLoh.circos']}
-                width={700}
-              />
-            </div>
-          )}
+          <div className="copy-number__circos">
+            <Image
+              image={images.find((img) => img.key === 'cnvLoh.circos')}
+              width={700}
+            />
+          </div>
           {groupedCnvs && (
             <>
               {Object.entries(groupedCnvs).map(([key, value]) => (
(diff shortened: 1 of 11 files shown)


elewis2 (approved):

app/views/ReportView/components/Expression/index.jsx
```diff
@@ -182,7 +182,7 @@ const Expression = () => {
               ))}
             </Paper>
           )}
-          {comparators && !Boolean(comparators.length) && (
+          {comparators && !comparators.length && (
```
kttkjl, line 185: unrelated to this line

on line 34 I believe it should be initialized to  `const [expOutliers, setExpOutliers] = useState([])`;

LinearProgress stays when expOutliers stay `undefined`

app/views/ReportView/components/ExpressionCorrelation/index.tsx
```diff
@@ -10,7 +10,7 @@ import ReportContext from '@/context/ReportContext';
 import ImageService from '@/services/reports/image.service';
 import DataTable from '@/components/DataTable';
 import { getPairwiseExpressionCorrelation } from '@/services/reports/pairwise-expression';
-import Image from '@/components/Image';
+import Image, { ImageType } from '@/components/Image';
```
kttkjl, line 13: I get what this is doing, but I don't think the API is updated to give an `array` instead of `map`

image

app/views/ReportView/components/Immune/index.tsx
```diff
@@ -52,16 +52,16 @@ const Immune = (): JSX.Element => {
         titleText="Immune Cell Types"
         canViewDetails
       />
-      {Object.values(images).length > 0 && (
+      {Boolean(images.length) && (
         <div>
           <Image
-            image={images['cibersort.combined_t-cell_scatter']}
+            image={images.find((img) => img.key === 'cibersort.combined_t-cell_scatter')}
```
kttkjl, line 58: could probably shorten these to `.find(({key}) => key === 'cibersort.combined_t-cell_scatter'`

app/views/ReportView/components/PathwayAnalysis/index.tsx
```diff
@@ -49,7 +49,7 @@ const PathwayAnalysis = ({
             `/reports/${report.ident}/image/retrieve/pathwayAnalysis.legend`,
             {},
           ).request();
-          setLegend(legendResp['pathwayAnalysis.legend']);
+          setLegend(legendResp[0]);
```
kttkjl, line 52: is there a record I can see this in action?

kttkjl (changes requested):
Code looks mostly fine, just some comments

Only glaring thing is that the images aren't coming back as arrays

app/views/ReportView/components/CopyNumber/index.tsx
```diff
@@ -128,14 +128,12 @@ const CopyNumber = (): JSX.Element => {
...
-            <div className="copy-number__circos">
-              <Image
-                image={images['cnvLoh.circos']}
-                width={700}
-              />
-            </div>
-          )}
+          <div className="copy-number__circos">
+            <Image
+              image={images.find((img) => img.key === 'cnvLoh.circos')}
```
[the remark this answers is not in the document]
kttkjl: awaiting on `https://github.com/bcgsc/pori_ipr_api/pull/54`, looks to be array returning

app/views/ReportView/components/Expression/index.jsx
Anna-log7: Line 62 should handle this case. I won't change this here if it's working since it's unrelated to the pr

app/views/ReportView/components/PathwayAnalysis/index.tsx
Anna-log7: Any report on the page `/pathway-analysis` should show this

app/views/ReportView/components/Expression/index.jsx
kttkjl: ah, it seems like this was due to the API problem 

```
Uncaught (in promise) TypeError: images.find is not a function
```

app/views/ReportView/components/PathwayAnalysis/index.tsx
kttkjl: looks to be an array return ... since there's no custom legends and this returns the expected type it's fine for now

kttkjl (approved):

PR #62, opened 2021-06-01
Add pharmacogenomic section to genomic report

Anna-log7:
-- not to be included in the coming release --
This is to add a new section called pharmacogenomic to the genomic report under the germline header. `/pharmacogenomic`


elewis2 (dismissed):

app/views/ReportView/index.jsx
```diff
@@ -213,6 +214,12 @@ const ReportView = () => {
                   )}
                   path={`${path}/immune`}
                 />
+                <Route
+                  render={(routeProps) => (
+                    <Pharmacogenomic {...routeProps} print={false} />
+                  )}
+                  path={`${path}/pharmacogenomic`}
+                />
```
kttkjl, line 222: non blocker, it doesn't look like `routeProps` is being used in Pharmacogenomic here (it's just to conform to how it's set up)?
but a new way of using Route without the `render` shenanigans is 

```
<Route path={`${path}/pharmacogenomic`}>
  <Pharmacogenomic print={false} />
</Route>
```

and then use hooks inside the component to get the route params you want so it's more clear in the actual component what you need from the router

kttkjl (changes requested):
types would be nice, but code looks fine 

I've been told there was no report that had pharmacogenomic data on it; tried to search id based on ticket, no cake

kttkjl (approved):

elewis2 (approved):

PR #67, opened 2021-06-07
Release/v6.3.1

Anna-log7:
IPR Client v6.3.1 Release Notes

Bugfixes:
- Fixed scrollbars not appearing on firefox
- Fixed an api request being sent with no data
- DEVSU-1542 - Fixed template api request
- DEVSU-1544 - Re-add WPC plot

### DIFF
--- a/app/views/ReportView/index.scss
+++ b/app/views/ReportView/index.scss
@@ -6,8 +6,6 @@
     height: 100%;
     display: flex;
     flex-direction: row;
-    min-height: -webkit-fill-available;
-    min-height: -moz-available;
   }
 
   &__content {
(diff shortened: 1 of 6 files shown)


app/views/ReportView/index.scss
```diff
@@ -3,10 +3,11 @@
 
 .report {
   &__container {
+    height: 100%;
     display: flex;
     flex-direction: row;
-    height: -webkit-fill-available;
-    height: -moz-available;
+    min-height: -webkit-fill-available;
+    min-height: -moz-available;
```
kttkjl, line 10: might not need min-height here if height is set to 100%?

kttkjl (approved):
height 100% fixes the problem on both firefox and chrome, not sure if the min-height is actually needed

app/views/ReportView/index.scss
Anna-log7: Removed

kttkjl (approved):

PR #71, opened 2021-06-17
Release/v6.3.2

Anna-log7:
### IPR Client v6.3.2 Release Notes

Bugfix
- DEVSU-1545 - Fixed erroring SV/CNV pages when no images are supplied
- DEVSU-1546 - Fixed index.html file being cached by SW
- Fixed scroll issues in germline reports

### DIFF
--- a/app/components/Image/index.tsx
+++ b/app/components/Image/index.tsx
@@ -6,7 +6,7 @@ import ImageType from './types.d';
 import './index.scss';
 
 type ImageProps = {
-  image: ImageType;
+  image?: ImageType;
   height?: number;
   width?: number;
   showTitle?: boolean;
--- a/app/components/Image/index.tsx
+++ b/app/components/Image/index.tsx
@@ -21,7 +21,7 @@ const Image = ({
     caption,
     format,
     key,
-  },
+  } = {} as ImageType,
   height = 0,
   width = 0,
   showTitle = false,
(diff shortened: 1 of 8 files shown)


elewis2 (dismissed):

elewis2 (dismissed):

app/components/Image/index.tsx
```diff
@@ -21,7 +21,7 @@ const Image = ({
     caption,
     format,
     key,
-  },
+  } = {} as ImageType,
```
kttkjl, line 24: Non-blocker but,

I read the ticket and tried those report ids, console didn't throw an error at me if I didn't set a default to the image

if the image is possibly undefined, then the type would probably have to change
```
type ImageProps = {
  image?: ImageType;
  ...
```

it's probably less work to destructure the image not in the props, but just like so 

```
const Image = ({
  image,
  height = 0,
  width = 0,
  showTitle = false,
  showCaption = false,
  isZoomable = true,
}: ImageProps): JSX.Element => {
  const {
    data, title, caption, format, key,
  } = image;

...
```

kttkjl (approved):

app/components/Image/index.tsx
Anna-log7: I'd still have to add a check for if the image exists in that example though right?

kttkjl: yep, I shoulda have illustrated it clearer

```
const { data, title, caption, format, key, } = image ?? {};
``` 
in this case for the default

kttkjl (approved):

kttkjl (approved):

PR #76, opened 2021-06-30
Feature/devsu 1547 add sw cache busting

Anna-log7:
This PR removes the registration of the service worker (reason explained on the ticket) and unregisters any active ones along with clearing any cached data by the SW.

I also added gzip compression to the build output to offset the speed loss of the SW removal, but the servers haven't enabled gzip serving yet (systems ticket is underway).

Finally, I changed the JWT library to something more lightweight as the old one imported the whole crypto library into the build


app/registerServiceWorker.js
```diff
@@ -28,18 +28,10 @@ function registerValidSW(swUrl) {
...
-              // available; please refresh." message in your web app.
-              console.log('New content is available; please refresh.');
-            } else {
-              // At this point, everything has been precached.
-              // It's the perfect time to display a
-              // "Content is cached for offline use." message.
-              console.log('Content is cached for offline use.');
-            }
+            // At this point, everything has been precached.
+            // It's the perfect time to display a
```
elewis2, line 32: is this text originally from a tutorial or something?

Anna-log7: Yes, it's the default text for a service worker

app/components/CacheBuster/index.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { useEffect } from 'react';
+
+const CacheBuster = ({
+  children,
+}) => {
+  useEffect(() => {
+    const delCaches = async () => {
+      const cacheKeys = await caches.keys();
```
kttkjl, line 8: caches is undefined in development mode -- think we can add a check here for `window.caches` ?

app/registerServiceWorker.js
```diff
@@ -28,18 +28,10 @@ function registerValidSW(swUrl) {
...
-              // It's the perfect time to display a "New content is
-              // available; please refresh." message in your web app.
-              console.log('New content is available; please refresh.');
-            } else {
-              // At this point, everything has been precached.
-              // It's the perfect time to display a
-              // "Content is cached for offline use." message.
-              console.log('Content is cached for offline use.');
-            }
+            // At this point, everything has been precached.
```
kttkjl, line 31: most of this code is just left here to unregister all "outdated" clients with the serviceWorkers?

and the section that is removed is not actually being used by anything right?

config/webpack/webpack.config.js
```diff
@@ -95,16 +96,19 @@ module.exports = {
...
     //   defaultSizes: 'gzip',
     //   excludeAssets: '.*\.hot-update\.js',
     // }),
   ],
   optimization: {
-    moduleIds: 'hashed',
+    runtimeChunk: 'single',
     splitChunks: {
       chunks: 'all',
+      minSize: 1000 * 500,
```
kttkjl, line 109: just curious, why 500k?

kttkjl (commented):
looks good save for the development env undefined check

config/webpack/webpack.config.js
Anna-log7: Trial and error testing so that quill wouldn't be loaded in more than one bundle as it's a fairly large package. I followed this article and it basically said use trial and error here https://blog.logrocket.com/guide-performance-optimization-webpack/

app/registerServiceWorker.js
Anna-log7: Yes that's correct. After this is released then I can remove this file entirely in a future update

app/components/CacheBuster/index.tsx
Anna-log7: Is it showing an error in the console? I'm not seeing anything

kttkjl: weird, probably because I'm not under SSL, are you? if you are then caches won't be undefined, you think that'd be a problem?

Anna-log7: I'm not on SSL, no. I looked it up though and you're right so I'll add some logic for if it's undefined

kttkjl (approved):

PR #79, opened 2021-07-13
Feature/devsu 1565 add confirmation to delete slides

Anna-log7:
Enhancement for confirming deletion of a slide before it's deleted. Normally there would be confirmations, but the slides API route doesn't un-sign a report like other routes.


kttkjl (approved):

kttkjl:
Non blocker, but it would be nice to see progress on delete when the operation is in progress.

I don't believe the dialog should disappear before the call finishes 

In terms of what the ticket describes, this PR does it

PR #86, opened 2021-07-27
Release/v6.4.1

Anna-log7:
### IPR Client v6.4.1 Release Notes

Bugfix:
- DEVSU-1588 - Fixed a bug where svgs would have erratic styling


Anna-log7:
Note: the regex change is so the regex only matches the `<svg>` attrs and not style attrs from other tags

elewis2 (approved):

kttkjl (approved):

PR #90, opened 2021-08-04
Bugfix/devsu 1590 fix ai svg regex

Anna-log7:
I've fixed the regex so that it matches correctly. Here's a breakdown of what's happening:

1. `(?<=<svg[^>]*?)` There should be `<svg` without `>` and then any chars (lazy) before the main capture group
2. `style=` Matches `style=`
3. `(\\?"[^"]*?\\?")` Captures (optional escaped) `"`, then as many characters before the next (optionally escaped) `"`
4. `|` OR
5. `(\\?'[^']*?\\?')` Same as 3, but with `'` instead of `"`


kttkjl (dismissed):
it might be worthwhile to add test cases for this, and cover all possible input test cases

Anna-log7:
> 
> 
> it might be worthwhile to add test cases for this, and cover all possible input test cases

@kttkjl I've added some tests

app/components/SvgImage/__tests__/SvgImage.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    render(
+      <SvgImage
+        image={mockSVGWithXMLTag}
+      />,
+    );
+
+    expect(await screen.findByTestId('mock')).toBeInTheDocument();
+  });
+
+  test('It fails to render an SVG that is not formatted correctly', async () => {
```
kttkjl, line 55: I'm confused about these test cases, wouldn't they all render even if the regex replacer has nothing to replace? 

cause from the comments I'm expecting
`<svg data-testid="mock" style="enable-background:new 0 0 1052 744;" .....>`
to be formatted to 
`<svg data-testid="mock" enable-background="new 0 0 1052 744" ....>`

Anna-log7: Oops, forgot to add those cases for checking the style attribute.

Correct output though is a bit harder to test from what I looked into since I'd have to test the output across elements (unless you have any other suggestions for how to do this)

kttkjl: >Correct output though is a bit harder to test from what I looked into since I'd have to test the output across elements
Oh I just checked the regex in console with the test `.svg` file you sent me the other day, and the results were like my comment above

>test the output across elements
I am not sure what this means ...? it sounds like we should add a test case for the "across elements" situation

>Also it won't render if there's a style tag that's still formatted as style="", the page will crash and unmount React
an extra check + test case for `style=""`? this seems like an edge case

Anna-log7: > I am not sure what this means ...? it sounds like we should add a test case for the "across elements" situation
It means that if I pass an `<svg>` with other tags nested in it like `<path>` or `<g>` then those will be rendered as a tree in the DOM and it's hard to test the whole tree of what's rendered

> an extra check + test case for style="" then? this seems like an edge case
I don't think I explained this correctly. I meant that without the regex I've added in the component then the page will crash when there's an svg with `style="any value here"`

kttkjl: >will crash when there's an svg with style="any value here"

that depends entirely on the data given to you right? We can just add more cases for "any value here" when it pops up

>It means that if I pass an <svg> with other tags nested in it like <path> or <g> then those will be rendered as a tree in the DOM and it's hard to test the whole tree of what's rendered

I think you've gotten a good idea for a setup on `mockSVGStyleInOtherTags` ... since you can just test for
- style attribute in non-svg tags to be unchanged
- style attribute in svg tags to be your expected output, or exists, something like `toHaveAttribute`

I'm not an expert on svgs ... do they have a weird rendering cycle that I'm not aware of that gets in the way of writing these tests?

kttkjl: as per offline discussion, the goal was to test that the svg would not render if the replacer regex didn't work correctly

kttkjl (approved):

PR #91, opened 2021-08-06
Bugfix/devsu 1595 germline report access

Anna-log7:
The main changes in this PR were to change access a bit in the client:
* External users should only be able to see reviewed/archived reports
* External users should not be able to access germline pages
* Hooks were added to replace the old `acl.service.js` file


app/views/ReportsView/index.tsx
```diff
@@ -25,19 +26,20 @@ const ReportsTableComponent = (): JSX.Element => {
...
   } = useGrid();
 
-  const { userDetails, adminUser } = useContext(SecurityContext);
+  const { adminAccess } = useResource();
+  const isExternalMode = useExternalMode();
   const [rowData, setRowData] = useState<ReportType[]>();
 
   useEffect(() => {
-    if (!rowData) {
+    if (isExternalMode !== undefined) {
```
kttkjl, line 34: should `rowData` check be removed here ...?

Anna-log7: No it should not, fixing that

kttkjl (approved):

PR #96, opened 2021-09-10
Bugfix/devsu 1611 variant edit fix

Anna-log7:
This fixes a bug where the gene couldn't be changed in the events section of the TGR. I've added an autocomplete textbox that pulls the available genes from the api so that the gene ident can be sent on change.


kttkjl (changes requested):
There is a bug on the autocomplete for local version where typing into the autocomplete for Gene doesn't seem to change the state of the textbox

Seems like the clear button doesn't work properly as well

kttkjl:
textboxState
Bug in question

kttkjl:
I think you can just remove onInputChange altogether, since the API grabs all of the Genes at the start anyways

Anna-log7:
> 
> 
> I think you can just remove the interaction with `onInputChange` altogether, since the API grabs all of the Genes at the start anyways
> noInputChange
> 
> Think the API needs update/change because in the original ticket, the target gene to be changed to is `AMER1`?

@kttkjl I see what you mean. I think in this case the report will need to be re-uploaded as the FAM123B gene is present in the rest of the sections as well.

kttkjl (approved):
bug fixed, looks good

PR #97, opened 2021-09-13
Merge changes to dev

Anna-log7:
merge back to dev


kttkjl (approved):
Is there something to setup on Github that auto merges things back to dev once master gets updated? Since like this doesn't -really- need any sort of review on the devs

Anna-log7:
> 
> 
> Is there something to setup on Github that auto merges things back to dev once master gets updated? Since like this doesn't -really- need any sort of review on the devs

Possibly, but out restrictions are such that we can't merge straight from master -> dev and there needs to be an intermediate branch. That may complicate things

PR #99, opened 2021-09-14
Feature/devsu 1178 sanitize comments html

Anna-log7:
- Added sanitization to comments to remove `img` tags and disallow certain links
- `style` tags will still work even though there was concern on the original ticket

This fixes potential XSS attacks that quill is vulnerable to through `img` tags and links. For links I've enable only free text (no http/https/mailto: etc) which will still allow for the links to GKB to exist.


app/views/ReportView/components/AnalystComments/__tests__/AnalystComments.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+};
+
+jest.mock('@/services/api');
+
+describe('AnalystComments', () => {
+  beforeEach(() => resetAllWhenMocks())
+
+  test('Img tags are sanitized', async () => {
+    when(api.get)
+      .mockImplementation(() => ({ request: async () => null }))
```
kttkjl, line 30: TS be screaming at you with the wrong typedefs, else this works 

I guess you can cheat with 
`when(api.get as (endpoint: string) => Partial<ApiCall>)`

kttkjl (changes requested):
does indeed sanitize before rendering, but has wrong types

app/views/ReportView/components/AnalystComments/__tests__/AnalystComments.test.tsx
Anna-log7: That works

kttkjl (approved):

PR #102, opened 2021-09-20
Bugfix/devsu 1631 fix admin endpoints

Anna-log7:
- Fixed endpoints sending the user object instead of ident to API
- Fixed select boxes for groups and projects


kttkjl (approved):
I can't access the Admin page, but code-wise the code looks correct

Perhaps some screenshots on what it is supposed to do would help

PR #110, opened 2021-10-28
Change size of immune images

Anna-log7:
* Changed the size of the immune images
* Changed the order of the cibersort immune images
* Changed the layout of the page so the images would display beside each other if there is space


app/views/ReportView/components/Immune/index.tsx
```diff
@@ -10,6 +10,15 @@ import withLoading, { WithLoadingInjectedProps } from '@/hoc/WithLoading';
 import { hlaColumnDefs, cellTypesColumnDefs } from './columnDefs';
 import { ImmuneType, HlaType } from './types';
 
+import './index.scss';
+
+const IMAGE_KEYS = [
+  'cibersort.combined_t-cell_scatter',
+  'cibersort.cd8_positive_t-cell_scatter',
+  'mixcr.circos_trb_vj_gene_usage',
+  'mixcr.dominiance_vs_alpha_beta_t-cells_scatter',
```
elewis2, line 19: is this the right mixcr name? dominiance, not dominance?

elewis2 (changes requested):
check mixcr image name

app/views/ReportView/components/Immune/index.tsx
Anna-log7: fixed

elewis2 (approved):

kttkjl (approved):

PR #111, opened 2021-10-28
Feature/devsu 1607 type api returns

Anna-log7:
This PR adds types to the api module using generics. I've also typed each call with the expected data type to be returned from the API.
Additionally, I removed usage of `ApiCallSet` except in cases where there needs to be a signature check. I did this because typing `ApiCallSet` is pretty tricky and I'm not sure it's possible to do in a descriptive way in typescript when there's various types in an array. If anyone has a solution for that though then I'd be happy to implement it.


kttkjl (approved):
This PR does what this PR describes itself to do, but the ApiCallSet's type `any[]` would need a fix down the line, along with multiple other type errors

kttkjl (approved):

PR #113, opened 2021-11-03
Feature/devsu 1637 upgrade mui

Anna-log7:
I've upgraded MUI to v5 in this PR by running the codemod on their site: https://mui.com/guides/migration-v4/
The build will fail because node 10 is still a task (that we need to remove). This PR won't be included in the 6.6.0 release


app/components/DataTable/components/ActionCellRenderer/index.jsx
```diff
@@ -78,131 +78,129 @@ function ActionCellRenderer(params) {
     setAnchorEl(null);
   };
 
-  return (
-    <>
-      {canViewDetails && (
+  return <>
```
kttkjl, line 81: think auto fix messed the styling on this

app/components/DataTable/components/GeneCellRenderer/index.jsx
```diff
@@ -16,45 +16,43 @@ function GeneCellRenderer(params) {
...
   const [showGeneViewer, setShowGeneViewer] = useState(false);
 
-  return (
-    <>
-      {value && value.split(/\s*::\s*|,\s/).map((val, index) => (
-        <React.Fragment key={index}>
-          {index > 0 && (
-            <span>
-              {value.includes(' :: ') ? ' :: ' : ', '}
+  return <>
```
kttkjl, line 19: here as well, auto fix messed with styling

app/views/ReportView/components/MutationBurden/components/TabCards/index.tsx
```diff
@@ -169,72 +169,70 @@ const TabCards = ({
...
-                    )}
-                  </Card>
-                )}
-              </React.Fragment>
-            );
-          })}
-      </div>
-    </>
-  );
+  return <>
```
kttkjl, line 172: auto styling problem

app/components/Sidebar/index.tsx
```diff
@@ -210,7 +210,7 @@ const Sidebar = (): JSX.Element => {
           {drawer}
         </Drawer>
       </Hidden>
-      <Hidden xsDown>
+      <Hidden smDown>
```
kttkjl, line 213: it doesn't look like xs v4 is diff from xs v5? why was this change here

kttkjl (changes requested):

app/components/Sidebar/index.tsx
Anna-log7: Not sure, I'll look into it

Anna-log7: https://mui.com/guides/migration-v4/#theme there was changes, but I don't know if they necessarily apply here. `xsDown` seems to mean less than 0px, so I think this change is correct. I tested it and `smDown` works for the behaviour I want.

app/views/ReportView/components/MutationBurden/components/TabCards/index.tsx
Anna-log7: fixed all of these

elewis2 (approved):

kttkjl (approved):

PR #120, opened 2021-11-16
DEVSU-1683 - account for type array for externalSource in CivicCellRenderer

kttkjl:
title

DEVSU-1683


kttkjl:
making this into a hotfix instead

PR #122, opened 2021-11-18
fix: apply changes done to Expression to SmallMut, CNV, SV

kttkjl:
- Add colDef for tpm and kIQR for SM, CNV, SV
- Add interactions to show either rpkm/tpm

DEVSU-1686

### DIFF
--- a/app/views/ReportView/components/CopyNumber/columnDefs.ts
+++ b/app/views/ReportView/components/CopyNumber/columnDefs.ts
@@ -1,4 +1,7 @@
-const columnDefs = [{
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { ColDef } from '@ag-grid-community/core';
+
+const columnDefs: ColDef[] = [{
   headerName: 'Gene',
   cellRenderer: 'GeneCellRenderer',
   cellRendererParams: { link: true },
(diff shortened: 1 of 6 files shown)


app/views/ReportView/components/CopyNumber/columnDefs.ts
```diff
@@ -34,17 +37,28 @@ const columnDefs = [{
...
-  field: 'gene.expressionVariants.rpkm',
-  hide: false,
+  colId: 'rpkm',
+  valueGetter: 'data.gene.expressionVariants.rpkm',
+  hide: true,
 }, {
   headerName: 'Expression (Normal FC)',
   colId: 'primarySiteFoldChange',
-  field: 'gene.expressionVariants.primarySiteFoldChange',
+  valueGetter: 'data.gene.expressionVariants.primarySiteFoldChange',
```
Anna-log7, line 46: Similar to TPM vs RPKM we want to show Expression (Normal FC) OR Expression (kIQR) depending on what's available

app/views/ReportView/components/CopyNumber/index.tsx
```diff
@@ -26,8 +26,8 @@ const TITLE_MAP = {
   lowExp: 'Lowly Expressed Tumour Suppressors with Copy Losses',
 };
 
-const getInfoDescription = (relevance: string) => `Copy variants where the variant matched 1 or 
-more statements of ${relevance} relevance in the knowledge base matches section. Details on these 
+const getInfoDescription = (relevance: string) => `Copy variants where the variant matched 1 or
```
Anna-log7, line 29: Removing whitespace here affects the string itself so it should be kept

Anna-log7 (changes requested):

app/views/ReportView/components/CopyNumber/index.tsx
kttkjl: must of been auto format

app/views/ReportView/components/CopyNumber/columnDefs.ts
kttkjl: just for copy number or for all 3 of these?

app/views/ReportView/components/CopyNumber/index.tsx
```diff
@@ -88,14 +89,32 @@ const CopyNumber = ({
...
                 break;
               }
               if (rpkm !== null) {
-                setVisibleCols((prevVal) => [...prevVal, 'rpkm']);
+                nextVisible.push('rpkm');
                 break;
               }
             }
+            /* Show either Normal or kIQR columns based on which is populated */
+            for (const {
```
Anna-log7, line 101: Some of these can have a lot of entries so it would be ideal to have the logic here in one loop instead of 2. Possibly something like:

```
            for (const {
              gene: {
                expressionVariants: {
                  tpm, rpkm, primarySiteFoldChange, primarySitekIQR
                },
              },
            } of cnvsResp) {
              /* Show either RPKM or TPM columns based on which is populated */
              if (tpm !== null && !nextVisible.includes('tpm')) {
                nextVisible.push('tpm');
              }
              if (rpkm !== null && !nextVisible.includes('rpkm')) {
                nextVisible.push('rpkm');
              }
              if (primarySiteFoldChange !== null && !nextVisible.includes('primarySiteFoldChange')) {
                nextVisible.push('primarySiteFoldChange');
              }
              if (primarySitekIQR !== null && !nextVisible.includes('primarySitekIQR')) {
                nextVisible.push('primarySitekIQR');
              }
              if (nextVisible.length === 2) {
                break;
              }
            }
```

Anna-log7 (changes requested):

app/views/ReportView/components/CopyNumber/index.tsx
kttkjl: yeah this will work better

Anna-log7 (approved):

PR #131, opened 2021-12-07
[DEVSU-1694] bugfix AlertDialog

kttkjl:
- Added additional handling in AlertDialog component
- Updated type of AlertDialogProps

DEVSU-1694

### DIFF
--- a/app/components/AlertDialog/__tests__/AlertDialog.test.tsx
+++ b/app/components/AlertDialog/__tests__/AlertDialog.test.tsx
@@ -1,0 +1,114 @@
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+import AlertDialog from '..';
+
+describe('AlertDialog', () => {
+  let component;
+
+  test('It renders', () => {
+    component = render(
+      <AlertDialog
+        isOpen={true}
+        onClose={() => { }}
+        title='Dialog Title'
+        text='Mock dialog text'
+      />
+    );
+    const { getByText } = component;
+    expect(getByText('Dialog Title')).toBeInTheDocument()
+    expect(getByText('Mock dialog text')).toBeInTheDocument()
+  });
+
+  test('It allows custom confirm and cancel text', () => {
+    component = render(
+      <AlertDialog
+        isOpen={true}
+        onClose={() => { }}
+        title='Dialog Title'
+        text='Mock dialog text'
+        confirmText='yas'
+        cancelText='nyet'
+      />
+    );
+    const { getByText } = component;
+    expect(getByText('yas')).toBeInTheDocument()
+    expect(getByText('nyet')).toBeInTheDocument()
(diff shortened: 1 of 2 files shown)


elewis2 (approved):

areisle (changes requested):

app/components/AlertDialog/index.tsx
```diff
@@ -30,8 +31,16 @@ const AlertDialog = ({
...
+    if (typeof evt === 'boolean') {
+      onClose(evt);
+      return;
+    }
+    onClose(false);
+  }, [onClose])
+
   return (
-    <Dialog open={isOpen} onClose={onClose}>
+    <Dialog open={isOpen} onClose={handleClose}>
```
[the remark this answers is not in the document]
kttkjl: Ah, I have been confused by the part by the callback passed in from parent components, good catch

elewis2 (dismissed):

app/components/AlertDialog/__tests__/AlertDialog.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+        cancelText='nyet'
+      />
+    );
+    const { getByText } = component;
+    expect(getByText('yas')).toBeInTheDocument()
+    expect(getByText('nyet')).toBeInTheDocument()
+  })
+
+
+  test('It calls onClose with the correct params', () => {
```
areisle, line 41: I would split this into 2 tests, one for returns true for confirmed when confirm is clicked and one for returns false for confirmed when cancel is clicked

app/components/AlertDialog/__tests__/AlertDialog.test.tsx
```diff
@@ -0,0 +1,100 @@
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+import AlertDialog from '..';
+
+jest.mock('@/services/api');
+
+describe('AlertDialog', () => {
+  let component = render(
```
areisle, line 8: this would be better inside the test. it makes it clearer what test this is for, and it's only used by the one test anyway

app/components/AlertDialog/__tests__/__snapshots__/AlertDialog.test.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`AlertDialog It renders 1`] = `<DocumentFragment />`;
```
areisle, line 3: I think I wouldn't bother with a snapshot in this case since this isn't including a snapshot of the dialog content anyway

areisle (commented):
like the added tests :) just a couple comments about readability and the snap file

areisle (approved):

PR #132, opened 2021-12-07
[DEVSU-1650] Export Germline report

kttkjl:
- Add check to filter out empty headernames and action column on exported files
- Updated actions on page to have dropdown menu like other pages
- Updated handleConfirmDelete to handleClose

DEVSU-1650

### DIFF
--- a/app/views/GermlineView/components/Report/index.tsx
+++ b/app/views/GermlineView/components/Report/index.tsx
@@ -1,13 +1,16 @@
-import React, { useEffect, useState, useCallback } from 'react';
+import React, { useEffect, useState, useCallback, useRef } from 'react';
 import { useParams, useHistory } from 'react-router-dom';
 import {
-  Button,
   Checkbox,
   FormControlLabel,
   Typography,
+  IconButton,
+  Menu,
+  MenuItem,
+  DialogProps,
 } from '@mui/material';
 import { AgGridReact } from '@ag-grid-community/react';
-
+import MoreHorizIcon from '@mui/icons-material/MoreHoriz';
 import useGrid from '@/hooks/useGrid';
 import api from '@/services/api';
 import GermlineReportContext from '@/context/GermlineReportContext';
@@ -20,7 +23,7 @@ import StrikethroughCell from './components/StrikethroughCell';
 import EditDialog from './components/EditDialog';
 import Reviews from './components/Reviews';
 import columnDefs from './columnDefs';
-
+import { getDate } from '@/utils/date';
 import './index.scss';
 
 type GermlineReportProps = WithLoadingInjectedProps;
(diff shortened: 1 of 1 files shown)


elewis2 (approved):

app/views/GermlineView/components/Report/index.tsx
```diff
@@ -98,8 +103,10 @@ const GermlineReport = ({
     />
   ), [onEdit]);
 
-  const handleConfirmDelete = useCallback(async (confirm) => {
+  const handleClose = useCallback(async (confirm) => {
     setShowAlertDialog(false);
+    setMenuAnchor(null);
+    // This could be an event or boolean
```
kttkjl, line 109: unnecessary comment, will remove

app/views/GermlineView/components/Report/index.tsx
```diff
@@ -98,8 +103,10 @@ const GermlineReport = ({
     />
   ), [onEdit]);
 
-  const handleConfirmDelete = useCallback(async (confirm) => {
+  const handleClose = useCallback(async (confirm) => {
```
areisle, line 106: non-blocking: given there's a couple popups in this component, I might make this name more verbose so it's clearer without looking where it's called which popup it's for. maybe `handleCloseDeleteAlert` or similar

app/views/GermlineView/components/Report/index.tsx
```diff
@@ -111,6 +118,34 @@ const GermlineReport = ({
...
+      columnSeparator: '\t',
+      columnKeys: colApi.getAllDisplayedColumns()
+        .filter((col) => col.colDef.headerName !== 'Actions' && col.colDef.headerName)
+        .map((col) => col.colId),
+      fileName: `ipr_${report.patientId}_${report.ident}_germline_${date}.tsv`,
+      processCellCallback: (({ value }) => (typeof value === 'string' ? value?.replace(/,/g, '') : value)),
+    });
+  }, [colApi, gridApi, report]);
+
+  const handleMenuItemClick = useCallback((action) => {
```
areisle, line 135: since this is only used by the export this function can be removed and the relevant logic of `setMenuAnchor(null);` added to  `handleTSVExport` instead

areisle (changes requested):
I still have to spin up the app to test the new feature, but think adding a test for the download functionality would be a good idea

app/views/GermlineView/components/Report/index.tsx
```diff
@@ -111,6 +118,34 @@ const GermlineReport = ({
...
+    switch (action) {
+      case 'delete':
+        console.log('Delete')
```
kttkjl, line 138: this .. was a mistake

kttkjl:
> I still have to spin up the app to test the new feature, but think adding a test for the download functionality would be a good idea

Since the export functionality is built into AgGrid itself, I guess the only test we'd do would be 
- Separate export function out into its own file
- Get output from AgGrid
- Check for all modifications needed from requirements on said output?

areisle:
adding tests for this turned out to be much more involved as the repo isn't really setup well for it yet, so I think that can be in a later PR

areisle (approved):

areisle (approved):

PR #135, opened 2021-12-31
Feature/devsu 1699 babelrc 7 mui 5

kttkjl:
Detailed in the ticket, the babel plugins were intended to minimize package size.

This is just updated to do the same as intended in the project, according to their docs


.babelrc
```diff
@@ -1,73 +1,71 @@
...
-        ["transform-imports", {
-          "@material-ui/core": {
-            "transform": "@material-ui/core/esm/${member}",
-            "preventFullImport": true
-          },
-          "@material-ui/icons": {
-            "transform": "@material-ui/icons/esm/${member}",
-            "preventFullImport": true
+        [
+          "babel-plugin-direct-import",
```
areisle, line 18: did you check the bundle size without this plugin? according to their docs it should now tree shake out of the box and this plugin should only be necessary for older webpack versions or if you're worried about start time in development

kttkjl: It doesn't look like it makes much difference in running dev instance asides from webpack taking 2x as long with the plugin 
image


However, orphan modules with the plugin is effectively cut in half in production build
image

areisle: _with_ the plugin? I thought the plugin was supposed to make it faster. interesting about the orphan modules

areisle: since this doesn't improve performance or affect bundle size, I think it would be better to ditch the plugin as it will make things simpler for future developers

kttkjl: yeah, agreed, doesn't seem to add much

areisle (approved):

PR #150, opened 2022-01-27
style: fix eslint config

areisle:
did not fix any of the linting errors found at this time
 to reduce the number of possible changes

they can be fixed in a future PR


kttkjl (commented):
Rules look good so far, would have to run this whenever node_module finishes doing what it needs...

kttkjl:
Asides from basic eslint errors, it seems like it's throwing a ton of circular-dependencies

areisle:
> Asides from basic eslint errors, it seems like it's throwing a ton of circular-dependencies

I followed one or 2 and these are correct, there is dependency cycles all over the place, most it seems because everything imports `@/services/api`, which imports `theme` from `./App` and `./App` imports all the views file. a lot of them would probably be fixed if theme was moved to it's own file

kttkjl:
> > Asides from basic eslint errors, it seems like it's throwing a ton of circular-dependencies
> 
> I followed one or 2 and these error messages are correct, there is dependency cycles all over the place, mostly it seems because everything imports `@/services/api`, which imports `theme` from `./App` and `./App` imports all the views file. A lot of the cycles would probably be fixed if theme was moved to it's own file
> 
> edit: I tried it, and it seems like theme in its own file fixes almost all the import cycle errors

Added ticket to track this fix https://www.bcgsc.ca/jira/browse/DEVSU-1718

kttkjl (approved):

kttkjl:
Thanks again Abbey!

PR #181, opened 2022-03-15
Feat/devsu 1757 appendix editor for admins

kttkjl:
Currently the appendix text is raw

image

due to the fact that on v25 of agGrid, auto height doesn't work properly on React components, so if we try to show the full preview on the Appendix Text column, it will only display a single line. However, the editor has all the formatting shown
image

on AgGrid [v26](https://www.ag-grid.com/changelog/?fixVersion=26.0.0) this problem is fixed, but that requires further testing due to breaking changes, so far we haven't used anything in the breaking changeslog, but I did encounter a core feature that was not documented ... but this is how it looks like with auto sizing, works great
image


kttkjl:
a test PR is setup [here](https://github.com/bcgsc/pori_ipr_client/pull/182) for v26 as well

kttkjl:
Hi @areisle this is the base PR for https://github.com/bcgsc/pori_ipr_client/pull/182

areisle:
so just making sure I'm understanding correctly, the last screenshot in the description is what you want it to eventually look like, but the first screenshot is what it looks like for this PR?

kttkjl:
> so just making sure I'm understanding correctly, the last screenshot in the description is what you want it to eventually look like, but the first screenshot is what it looks like for this PR?

That is correct yes, the last screenshot can be seen on the other PR that builds on top of this, I've tried to fix as many things that came up from upgrading to agGrid 26 from 25 as I can see in that PR.

areisle:
if the reason for updating to 26 is just to fix the height for this particular column, I don't think it's necessary. since you're trying to render html from what I can see, you can use the builtin renderer for the html string without resorting to a react component.
adding:
```
cellRenderer: ({ value }) => value || ''
```
on the colDef will render the value as html
and then autoHeight works since it's not using react
https://plnkr.co/edit/Tq3gZrzXRsuHWpxS?preview

kttkjl:
>if the reason for updating to 26 is just to fix the height for this particular column
Partially, yes it is to set the height properly, but autoheight was broken for react components in 25 so I figured to upgrade anyways

But since you showed that example (cool, didn't know you can straight up render) I guess upgrading to v26 isn't needed here

kttkjl:
Will fix to just the renderer with raw text then

kttkjl:
hmmm, seem like the autoHeight/wrapText still has a problem, regardless whether or not this is a react component
image

app/views/AdminView/components/Appendices/columnDefs.ts
```diff
@@ -0,0 +1,36 @@
...
+const columnDefs: ColDef[] = [
+  {
+    headerName: 'Name',
+    field: 'name',
+    hide: false,
+  },
+  {
+    headerName: 'Org',
+    field: 'organization',
+    hide: false,
```
areisle, line 13: why are you setting `hide: false`?

app/views/AdminView/components/Appendices/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const [appendices, setAppendices] = useState<TemplateWithAppendix[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [isEditing, setIsEditing] = useState(false);
+  const [editingData, setEditingData] = useState(null);
+
+  // Grab templates
+  useEffect(() => {
+    const getTemplates = async () => {
+      const templatesResp = await api.get('/templates').request();
+      setTemplates(templatesResp);
```
areisle, line 26: this will cause an setState on unmounted component error. this can be fixed by adding a cancelled flag, but I think better would be to just add `react-query` to the project and use that

app/views/AdminView/components/Appendices/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+    };
+    getTemplates();
+  }, []);
+
+  // Grab template appendices
+  useEffect(() => {
+    const templateIdToObj = {};
```
areisle, line 33: similarly this hook also needs a cancelled flag

app/views/AdminView/components/Appendices/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+    <>
+      {!isLoading && (
+        <>
+          <DataTable
+            columnDefs={columnDefs}
+            rowData={appendices}
+            titleText="Appendices"
+            canEdit
+            onEdit={handleOnEdit}
+
```
areisle, line 106: extra whitespace here

app/views/AdminView/components/Appendices/columnDefs.ts
```diff
@@ -0,0 +1,36 @@
+import { formatDate } from '@/utils/date';
+import { ColDef } from '@ag-grid-community/core';
+
+const columnDefs: ColDef[] = [
```
areisle, line 4: non-blocking, but since there's only a few columns so this isn't a massive array, I think it would be better just in the same file as the table

app/views/AdminView/components/Appendices/columnDefs.ts
```diff
@@ -0,0 +1,36 @@
...
+    hide: false,
+  },
+  {
+    headerName: 'Org',
+    field: 'organization',
+    hide: false,
+  },
+  {
+    headerName: 'Created',
+    valueGetter: ({ data }) => formatDate(data.createdAt),
```
areisle, line 17: this is getting cut off, so I think it would be good to have a fixed width for this column

app/components/Sidebar/index.tsx
```diff
@@ -195,6 +196,23 @@ const Sidebar = (): JSX.Element => {
                 </Typography>
               </Link>
             </ListItem>
+            <ListItem
```
areisle, line 199: I can see this is just copied from the existing pattern, so no need to do in this PR. but all of these list items have a lot of shared logic so I think it would be beneficial to extract that into a small component

app/views/AdminView/components/Appendices/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const handleOnEdit = useCallback((rowData) => {
+    setEditingData(rowData);
+    setIsEditing(true);
+  }, []);
+
+  const handleEditClose = useCallback(async (nextData) => {
+    if (nextData) {
+      const sanitizedText = sanitizeHtml(nextData);
+      if (!editingData?.appendix) {
+        const res = await api.post(`/templates/${editingData.ident}/appendix`, { text: sanitizedText }).request();
```
areisle, line 64: these 2 blocks share a lot of logic. I think it would be cleaner to combine them as follows:

```ts
const isNew = !editingData?.appendix
const res = await (isNew ? api.create : api.put)(`/templates/${editingData.ident}/appendix`, { text: sanitizedText }).request();

// Find appendix in state, update
setAppendices((currAppendices) => {
  const index = currAppendices.findIndex((app) => app.ident === editingData.ident);
  const nextAppendices = [...currAppendices];
  nextAppendices[index] = {
    ...nextAppendices[index],
    appendix: res,
  };
  return nextAppendices;
});

snackbar.success(`Appendix successfully ${isNew ? 'created' : 'updated'}.`);
setIsEditing(false);
```

areisle (changes requested):

app/views/AdminView/components/Appendices/columnDefs.ts
kttkjl: bad copy-paste, will remove

app/views/AdminView/components/Appendices/index.tsx
kttkjl: cancelled flag?

kttkjl: like so? 

```
    let cancelled = false;
    const getTemplates = async () => {
      const templatesResp = await api.get('/templates').request();
      if (!cancelled) {
        setTemplates(templatesResp);
      }
    };
    getTemplates();
    return function cleanup() { cancelled = true; };
    ```

app/views/AdminView/components/Appendices/index.tsx
kttkjl: Oh cool didn't know you can just ternary the function like that

app/components/Sidebar/index.tsx
kttkjl: https://www.bcgsc.ca/jira/browse/DEVSU-1788

app/views/AdminView/components/Appendices/index.tsx
kttkjl: [ef72650](https://github.com/bcgsc/pori_ipr_client/pull/181/commits/ef72650885d99aacf47b8fb03b3261fda8fa5e1a)

app/views/AdminView/components/Appendices/index.tsx
kttkjl: [ef72650](https://github.com/bcgsc/pori_ipr_client/pull/181/commits/ef72650885d99aacf47b8fb03b3261fda8fa5e1a)

areisle (approved):

PR #182, opened 2022-03-15
Feat/devsu 1757 gridv26 appendix editor for admins

kttkjl:
Test PR, more changes probably will be had due to breaking changes in agGrid 26


kttkjl:
oh hold until we get more time to work with v26

PR #183, opened 2022-03-16
fix: change TCGA headername for appendices

kttkjl:
DEVSU-1752


app/views/ReportView/components/Appendices/index.tsx
```diff
@@ -157,7 +157,7 @@ const Appendices = ({
                 <DataTable
                   columnDefs={tcgaAcronymsColumnDefs}
                   rowData={tcga}
-                  titleText="TCGA Acronyms"
+                  titleText="TCGA Acronyms for genomic report"
```
elewis2, line 160: Genomic Report should be capitalized

areisle (approved):

PR #187, opened 2022-03-18
feat: SV images now scale to view height

kttkjl:
- Add override to Image component to directly affect props in img tag

DEVSU-1137

### DIFF
--- a/app/components/Image/index.tsx
+++ b/app/components/Image/index.tsx
@@ -7,11 +7,13 @@ import './index.scss';
 
 type ImageProps = {
   image?: ImageType;
-  height?: number;
-  width?: number;
+  height?: number | string;
+  width?: number | string;
   showTitle?: boolean;
   showCaption?: boolean;
   isZoomable?: boolean;
+  /** Props to the img tag directly */
+  imgProps?: React.ComponentPropsWithoutRef<'img'>;
 };
 
 const Image = ({
@@ -22,11 +24,12 @@ const Image = ({
     format,
     key,
   } = {} as ImageType,
-  height = 0,
-  width = 0,
+  height = undefined,
+  width = undefined,
   showTitle = false,
   showCaption = false,
   isZoomable = true,
+  imgProps = {},
 }: ImageProps): JSX.Element => {
   const [isZoomed, setIsZoomed] = useState(false);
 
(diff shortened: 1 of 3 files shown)


areisle:
would you be able to post some before and after screenshots? I think it would make understanding the changes easier

kttkjl:
Before
image

After
vhmax

app/components/Image/index.tsx
```diff
@@ -7,11 +7,13 @@ import './index.scss';
...
   image?: ImageType;
-  height?: number;
-  width?: number;
+  height?: number | string;
+  width?: number | string;
   showTitle?: boolean;
   showCaption?: boolean;
   isZoomable?: boolean;
+  /** Props to the img tag directly */
+  imgProps?: React.ComponentPropsWithoutRef<'img'>;
```
areisle, line 16: I think it's confusing to call this `imgProps` when a lot of the other props also end up directly on the `img` tag (ex. width, height etc). Since what you actually want is just the maxHeight, I would just add that as the prop. or if you know there's other styles you're going to be adding maybe `imgStyles` as the prop

areisle: actually since the calculation for the `maxHeight` in the 2 cases where it's applied is the same, maybe a prop of like 
`shrinkToFitViewport={true}` might be better still

areisle (commented):
I tested the changes and the height does now scale with the viewport height, I just have some concerns about the name for the prop

app/components/Image/index.tsx
kttkjl: since there are other places where the `Image` component is used, my thinking is that if the normal height/width prop is not enough to style the component, then this prop would provide the override. Sure this is an encompass-all prop, but MUI does the same escape hatch thing as well, and if we do need more stuff in the future that's going to be more changes to the prop which IMO is just extra work 

`shrinkToFitViewport={true}` could be a prop, but it'd probably needs to specify `vw` or `vh` ...?

kttkjl: `imgStyles` would probably be our best bet here,

kttkjl: [923f6f4](https://github.com/bcgsc/pori_ipr_client/pull/187/commits/923f6f4402e53ceeb1fafcd75be8f833a7cc5868)

areisle: > shrinkToFitViewport={true} could be a prop, but it'd probably needs to specify vw or vh ...?

in both cases where you're setting maxHeight the code is identical so there would be no need to supply a value, just move that code into the Image component instead and apply conditionally

PR #191, opened 2022-03-25
[DEVSU-1764] VariantChips germline tag

kttkjl:
Added check to Variant Chips in displaying chips in cases where the variant is of type germline

Related to [this PR](https://github.com/bcgsc/pori_ipr_api/pull/144)

DEVSU-1764


app/views/ReportView/components/GenomicSummary/components/VariantChips/index.jsx
```diff
@@ -40,21 +40,62 @@ const VariantChips = (props) => {
...
     <div>
       <div>
         {Boolean(variants.length) && (
           <>
-            {variants.map(variant => (
+            {variants.map((variant) => (
               <React.Fragment key={variant.geneVariant}>
                 <Chip
-                  label={variant.geneVariant}
+                  label={`${variant.geneVariant} ${variant.germline ? '(germline)' : ''}`}
```
kttkjl, line 98: Just this line is the thing, everything else is just moving things around

kttkjl:
Not actual data, but this seem like a good report to show all the differences. For the sake of time, storybook would be in a later PR if anything to illustrate the different types of tags

image


image

app/views/ReportView/components/GenomicSummary/components/VariantChips/index.jsx
areisle: can skip for this one, but in future it would be helpful if changes that were just refactoring were split into their own commit

areisle (approved):

PR #218, opened 2022-05-10
[DEVSU-1799] add hyperlink option to appendix editor

kttkjl:
hyperlink option for appendix editor should be added for easier link editing 

(We could just straight up edit the HTML that is sent to the backend but that's more work)


elewis2:
So how will this work at present? Wouldn't the href tag get removed by the api?

kttkjl:
> So how will this work at present? Wouldn't the href tag get removed by the api?

in terms of SAVING the link into the backend, no this would not fix the problem, but in order for the user (admin) to add a link via the GUI, this change would be required

kttkjl:
@elewis2 this should be all the updates, the latest commit has `rel='noopener noreferrer'` described by systems

elewis2 (approved):

PR #230, opened 2022-06-10
[DEVSU-1693] fix: double window open for Reports link out

kttkjl:
- removed extra action handler onMouseUp
- used Link component for Icon

Screenshots depicting old and new interactions on original ticket


kttkjl:
Whoops, didn't run test, will update now

PR #247, opened 2022-08-02
[DEVSU-1829] collapse row web version

kttkjl:
Extension of PR #234 

Works in storybook
image

image


elewis2 (approved):

app/components/DataTable/components/ColumnPicker/index.tsx
```diff
@@ -77,7 +77,7 @@ const ColumnPicker = ({
                     onChange={(event) => handleChange(event, row?.getColId())}
                   />
                 )}
-                label={row.name}
+                label={row.name ? row.name : row.getColId()}
```
areisle, line 80: I realize this was already called this, but renaming `row` to `col` would be a less confusing variable name. also rather than using name/colId here I would use [`col.getDisplayNameForColumn()` ](https://www.ag-grid.com/javascript-data-grid/column-api/#reference-Visibility%20and%20Display-getDisplayNameForColumn) as that will match what they see in the column header

app/components/DataTable/index.scss
```diff
@@ -17,6 +17,20 @@
     .ag-row:last-child {
       border-bottom: none;
     }
+
+    .cell-span, .cell-span--last {
```
areisle, line 21: is this all just because you want it centered vertically vs the default top aligned or was there other issues with the borders?

app/components/DataTable/index.tsx
```diff
@@ -29,6 +30,65 @@ const MAX_VISIBLE_ROWS = 12;
 const MAX_TABLE_HEIGHT = '517px';
 const PAGE_TOP_OFFSET = 56 + 57;
 
+const getRowspanColDefs = (colDefs: ColDef[], displayedRows: RowNode[], colsToCollapse: string[]): ColDef[] => {
```
areisle, line 33: I think it would be good to have a docstring explaining what this is doing

app/components/DataTable/index.scss
kttkjl: both actually, there was the issue where the rowSpanned cell was not actually the last row according to the table, so there would be a double border at the visual bottom of the table

app/components/DataTable/components/ColumnPicker/index.tsx
kttkjl: oh cool, yeah I'll try that method

.storybook/main.js
```diff
@@ -9,7 +9,8 @@ module.exports = {
   ],
   addons: [
     "@storybook/addon-links",
-    "@storybook/addon-essentials"
+    "@storybook/addon-essentials",
+    "@react-theming/storybook-addon",
```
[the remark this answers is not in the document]
kttkjl: this was only added to display the MUI theme for the datatable, it looks different without ... and the other one doesn't support the new MUI version

app/components/DataTable/components/ColumnPicker/index.tsx
kttkjl: looks like current version of agGrid we have does not support `getDisplayNameForColumn`

app/components/DataTable/DataTable.stories.tsx
```diff
@@ -0,0 +1,155 @@
...
+};
+
+export default {
+  title: 'components/DataTable',
+  component: DataTable,
+};
+
+const Template = (args) => <DataTable {...args} />;
+
+export const Normal: Story<DataTableProps> = Template.bind({});
```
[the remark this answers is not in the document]
kttkjl: the newer method does make more sense than the current one, though we'll deal with it when storybook 7 comes out

.storybook/main.js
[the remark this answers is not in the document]
areisle: I'm not sure I understand. you don't need an addon to use the mui theme you can just write a decorator

app/components/DataTable/components/ColumnPicker/index.tsx
areisle: bummer

.storybook/main.js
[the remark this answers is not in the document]
kttkjl: mmm, tried it again with just the decorator and it seems to work fine, not sure why it didn't work the first time messing with it.

areisle (approved):

elewis2 (approved):

PR #248, opened 2022-08-05
[DEVSU-1751] tmbur to genomic

kttkjl:


kttkjl:
API shows a return of an actual object instead of an array of objects, will update PR to reflect

elewis2 (dismissed):

kttkjl:
all fields with TMB should have units of (mut/mb) now

elewis2 (approved):

PR #250, opened 2022-08-17
[DEVSU-1859] Unable to add specific variant in IPR therapeutic target table

kttkjl:
[this](https://github.com/bcgsc/pori_ipr_client/pull/250/files#diff-caeeb902ebeee40101faa337eccc43a26504bd256d967261d6ae25a22f3030a1R53) line is the line to look for, everything else is just adding typscript

### DIFF
--- a/app/views/ReportView/components/TherapeuticTargets/components/AutocompleteHandler/index.tsx
+++ b/app/views/ReportView/components/TherapeuticTargets/components/AutocompleteHandler/index.tsx
@@ -1,38 +1,36 @@
 import React, { useState, useEffect, useCallback } from 'react';
-import PropTypes from 'prop-types';
-import { Autocomplete } from '@mui/material';
 import {
   TextField,
   CircularProgress,
+  Autocomplete,
 } from '@mui/material';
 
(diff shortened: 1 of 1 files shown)


app/views/ReportView/components/TherapeuticTargets/components/AutocompleteHandler/index.tsx
```diff
@@ -96,20 +92,8 @@ const AutocompleteHandler = (props) => {
...
-AutocompleteHandler.defaultProps = {
-  defaultValue: '',
-  required: false,
-  onChange: () => { },
-  minCharacters: 3,
+export {
+  AutocompleteHandler,
 };
 
 export default AutocompleteHandler;
```
elewis2, line 99: are both exports needed?

kttkjl: not required no, but it is more of a convention thing, just that if we do decide to export more things from this module, we can be more modular about it

elewis2 (approved):

PR #253, opened 2022-08-18
[DEVSU-1751] tmbur addition #2

kttkjl:
move tmbur calls out from group call, since it can come back as 404, and makes sure it is silent failure


app/views/ReportView/components/GenomicSummary/index.tsx
```diff
@@ -313,11 +318,11 @@ const GenomicSummary = ({
         },
         {
           term: 'Genome SNV TMB (mut/mb)', // float
-          value: tmburMutBur?.genomeIndelTmb.toString(),
+          value: tmburMutBur ? tmburMutBur.genomeIndelTmb.toString() : '',
```
kttkjl, line 321: this change is that if there are no tmburMutBur data, this field would not even display

elewis2 (approved):

PR #263, opened 2022-09-14
[DEVSU-1856] deprecate msi from report

kttkjl:
Uses TMBur data instead of MSI data, if available 

This is assuming that all the newer reports WILL have TMBur data. If newer reports do not have TMBur data, then MSI would be displayed if it has data


kttkjl:
According to the ticket, some changes are needed, will start on it now

elewis2 (approved):

PR #271, opened 2022-10-21
[DEVSU-1871] adminview/user error handling

kttkjl:
- add snackbar messages to denote which step messed up

DEVSU-1871


elewis2:
I'm actually getting two popups when a user is successfully created or edited - a black one above a green one

elewis2 (changes requested):
There are two popups when a user record is created or edited - one black over one green. Probably only the green one is needed

kttkjl:
> There are two popups when a user record is created or edited - one black over one green. Probably only the green one is needed

Oh whoops, that wasn't taken into account -- will fix

Nithriel (approved):

elewis2 (approved):

PR #262, opened 2022-09-13
feature/DEVSU-1872-add-evidence-level-to-variants

Nithriel:
**Features**:
- Add ipr evidence level to variant tables

PR #272, opened 2022-10-21
[DEVSU-1873] add evidence level to variants

kttkjl:
Renamed branch, so a new PR based on @Nithriel 's original on #262 

Features:

Add ipr evidence level to variant tables


elewis2 (approved):

kttkjl:
Just making a note here, this change applies to the rapid report as well, since this is a template, and the rapid report uses these fields so it is not limited to genomic

PR #288, opened 2023-01-14
[DEVSU-1903] add reported in PCP-TGR field

kttkjl:
- chore: updated gridApi call accessors instead of direct access
- fix: resizing when "additional info" is shown -- previously table is a jumbled mess, now it properly renders correct boxes
- added feature of setting previously reported by checkbox (disabled for now until API supports it)

Screenshot as is, there is no report with "yes" or "no" so it's all blank for now
image

### DIFF
--- a/app/views/GermlineView/components/Report/columnDefs.ts
+++ b/app/views/GermlineView/components/Report/columnDefs.ts
@@ -175,12 +175,25 @@ const columnDefs = [
   {
     headerName: 'Additional Info',
     field: 'additionalInfo',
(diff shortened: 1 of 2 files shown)


elewis2 (approved):

elewis2 (approved):

kttkjl:
if the value is "null" in the backend, the field will return "not set" instead of "yes" or "no"

kttkjl:
image

elewis2 (approved):

PR #302, opened 2023-03-23
[DEVSU-1953] update Rapid Alt/Total title to reflect msi

kttkjl:


elewis2 (approved):

elewis2:
Actually, msi-type variants probably won't have anything in this field (neither will tmbur variants), it's the fusions/structural variants that will be coming in with tumour rna rather than dna (which is why we won't specify rna/dna).

PR #304, opened 2023-03-28
[DEVSU-1949] pmid links

kttkjl:
Fixes when the first in the list of items in PMIDs is NOT a pmid into rest just text

Further development regarding the whole field would be needed to deal with text and PMID, and which PMID should be the primary entrypoint

### DIFF
--- a/app/components/DataTable/components/ArrayCellRenderer/__tests__/ArrayCellRenderer.test.ts
+++ b/app/components/DataTable/components/ArrayCellRenderer/__tests__/ArrayCellRenderer.test.ts
@@ -58,7 +58,7 @@ describe('ArrayCellRenderer', () => {
 
     const elem = await screen.findByText(mockLinkArrayData.test[0]);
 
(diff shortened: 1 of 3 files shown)


kttkjl:
hmmm tests needs to be updated

kttkjl:
Quickest fix would be to remove the test for generating links if given half a link -- since how we're using `ArrayCellRenderer` does not deal with half-links such as `something.com` or `otherthing.net` ... and it also does not make sense in a design perspective that we would be prepending `http://` before a string to build said URL

Unless the PMID field have half-created links instead of just strings and pmids ... which in that case the PMID field would need a revamp

Long term solution would be separating the current `ArrayCellRenderer` into two components
- one that would assume the links given to it are actual links (no more prepending) 
- one that deals with array of string titles and pmids, and generate PMID links appropriately (after processing the array itself)

elewis2 (approved):

PR #318, opened 2023-05-05
[DEVSU-1971] detailDialog nested attributes display

kttkjl:
- added expandable array of objects as a feature


kttkjl:
tests prob broken because objects are defaulted to be collapsed

PR #321, opened 2023-05-11
[DEVSU-1959] kbmatch aggregation fix

kttkjl:
Added relevance and disease as unique key components to kbmatch tables aggregation

### DIFF
--- a/app/views/ReportView/components/KbMatches/coalesce.ts
+++ b/app/views/ReportView/components/KbMatches/coalesce.ts
@@ -1,3 +1,7 @@
+import {
+  KbMatchType, AnyVariantType,
+} from '@/common';
+
 class CoalesceEntriesError extends Error {
   constructor(error) {
     super(`Error coalescing entry: ${error.message}`);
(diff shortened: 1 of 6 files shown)


app/views/ReportView/components/KbMatches/coalesce.ts
```diff
@@ -11,56 +15,69 @@ class CoalesceEntriesError extends Error {
...
-      } = entry;
-      return `${gene1Name}${delimiter}${gene2Name}${delimiter}${context}${delimiter}${variantName}`;
+      } = entry as KbMatchType<'sv'>;
+      return `${gene1Name}${delimiter}${gene2Name}${delimiter}${commonSuffix}`;
     }
 
     if (entry.variantType === 'msi' || entry.variantType === 'tmb') {
-      const { kbCategory } = variant;
-      return `${kbCategory}${delimiter}${context}${delimiter}${variantName}`;
+      const { kbCategory } = variant as KbMatchType<'tmb' | 'msi'>['variant'];
```
elewis2, line 73: it's possible kbCategory will be null for some tmb results; will this break table display (ie cause the rest of the table not to load)?

elewis2 (changes requested):

app/views/ReportView/components/KbMatches/coalesce.ts
kttkjl: rest of the table would load -- it's just that this would concatenate with other possible rows that have `kbCategory === null` and the same `commonSuffix`, though if the unique identification for this type of variant is null, then perhaps we should change it to another field where it couldn't be null

elewis2 (approved):

PR #323, opened 2023-05-16
[DEVSU-1984] add captiv8 info to rapid report summary display

kttkjl:


kttkjl:
@elewis2 fixed the typing problem we were talkin about too

PR #330, opened 2023-05-26
[DEVSU-1992] add VAF% text display for rapid/therapeuticAssociation

kttkjl:


app/views/ReportView/components/RapidSummary/columnDefs.ts
```diff
@@ -68,7 +68,7 @@ const therapeuticAssociationColDefs = [
         return ((tumourAltCount / tumourDepth) * 100).toFixed(0);
       }
       if ((rnaAltCount && rnaDepth) || (rnaAltCount === 0 || rnaDepth === 0)) {
-        return `${rnaAltCount}/${rnaDepth}`;
+        return `N/A (RNA): ${((rnaAltCount / rnaDepth) * 100).toFixed(0)}`;
       }
```
elewis2, line 72: I believe that if it's RNA the analysts actually don't want the VAF calculated, so the entry in the VAF column should just be 'N/A (RNA)'

elewis2 (changes requested):

app/views/ReportView/components/RapidSummary/columnDefs.ts
elewis2: This change needs to be applied to the Cancer Relevance table, as well.

kttkjl: [981f0a5](https://github.com/bcgsc/pori_ipr_client/pull/330/commits/981f0a55089a46ead5980a8e204f1ed46df3a16d)

elewis2 (approved):

PR #332, opened 2023-06-02
[DEVSU-1967] change prio for checkAccess to prio allowlist

kttkjl:
DEVSU-1967

### DIFF
--- a/app/utils/__tests__/checkAccess.test.ts
+++ b/app/utils/__tests__/checkAccess.test.ts
@@ -1,0 +1,30 @@
+import checkAccess from '../checkAccess';
+
+describe('checkAccess', () => {
+  const mockUserGroups = [
+    {
+      name: 'clinician',
+      ident: 'uuid1',
+    },
+    {
+      name: 'manager',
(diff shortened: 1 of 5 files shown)


Nithriel (approved):

app/utils/__tests__/checkAccess.test.ts
```diff
@@ -0,0 +1,30 @@
...
+      name: 'clinician',
+      ident: 'uuid1',
+    },
+    {
+      name: 'manager',
+      ident: 'uuid2',
+    },
+  ];
+
+  test('Allow access when both allow and block list is empty', () => {
```
elewis2, line 15: does this mean a user can view data without having roles at all?

kttkjl: If the develop does not give an explicit `blockList`, but rather just undefined/null, then `checkAccess` defaults to blocking all. However, if the user provides a `[]` as the input for `blockList`, then it's open to everything

elewis2 (approved):

PR #335, opened 2023-06-16
[DEVSU-2005] Update IPR admin/projects to load project specific data on dialog edit

mathieulemieux:
https://www.bcgsc.ca/jira/browse/DEVSU-2005

### DIFF
--- a/app/views/ProjectsView/components/AddEditProjectDialog/index.scss
+++ b/app/views/ProjectsView/components/AddEditProjectDialog/index.scss
@@ -4,3 +4,6 @@
     padding: 0 16px 0 0;
   }
 }
+.edit-dialog__section-title {
+  margin-top: 8px;
+}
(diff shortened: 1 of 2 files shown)


app/views/ProjectsView/components/AddEditProjectDialog/index.scss
```diff
@@ -3,4 +3,7 @@
     margin: 8px;
     padding: 0 16px 0 0;
   }
+}
+.edit-dialog__section-title {
```
kttkjl, line 7: for future development, if `edit-dialog` gets more elements, we should split it via

```
.edit-dialog {
  &__section-title {
    ...
  }
  &__<some other element> {
    ...
  }
}
```

kttkjl (dismissed):
In terms of the ticket, the task to be done is done 

Personally would rename this title of PR to `[ticket-number] <purpose/title of ticket>`

Some further suggestions:

- Could disable the "submit" button until the reports API call is done
- Disable "submit" until the requests for updating the project is done (so they don't double submit thinking the submit button didn't do anything)
- Add an `abort()` call whenever this component unmounts in the callbacks to cancel (we don't have examples existing in the code, but the functionality exists)

kttkjl (approved):

PR #339, opened 2023-07-14
[DEVSU-2015] kctoken timeout notification

kttkjl:
Added a modal that starts timer when user logs in
  - Modal gives option to ignore or refresh token
 
Added listener for when user navigates the site to refresh token
Added error handling (notification) when editing Appendices errors out

### DIFF
--- a/app/services/management/auth.js
+++ b/app/services/management/auth.js
@@ -2,9 +2,9 @@ import Keycloak from 'keycloak-js';
 import jwtDecode from 'jwt-decode';
 
 const keycloak = Keycloak({
-  'realm': window._env_.KEYCLOAK_REALM,
-  'clientId': window._env_.KEYCLOAK_CLIENT_ID,
-  'url': window._env_.KEYCLOAK_URL,
+  realm: window._env_.KEYCLOAK_REALM,
+  clientId: window._env_.KEYCLOAK_CLIENT_ID,
+  url: window._env_.KEYCLOAK_URL,
 });
 
 const getReferrerUri = () => localStorage.getItem(CONFIG.STORAGE.REFERRER);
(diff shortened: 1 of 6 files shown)


app/services/management/auth.js
```diff
@@ -68,20 +68,23 @@ const login = async (referrerUri = null) => {
...
+      checkLoginIframe: false,
+    });
+  } else {
+    await keycloak.updateToken(Number.MAX_SAFE_INTEGER);
+  }
 };
 
 const logout = async () => {
   try {
-    await keycloak.init();
```
elewis2, line 82: what was this line doing?

app/views/MainView/index.tsx
```diff
@@ -29,6 +38,82 @@ const PatientsView = lazy(() => import('../PatientsView'));
...
+  setAuthorizationToken: React.Dispatch<React.SetStateAction<string>>;
+};
+
+const TimeoutModal = memo(({ authorizationToken, setAuthorizationToken }: TimeoutModalPropTypes) => {
+  const { location: { key: locationKey } } = useHistory();
+  const [open, setIsOpen] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+  const timerRef = useRef(null);
+
+  // Refresh token everytime user navigates the page, much cleaner than if they did ANY action
```
elewis2, line 55: makes sense, but does it slow things down detectably?

app/views/ReportView/components/AnalystComments/index.tsx
```diff
@@ -99,40 +99,45 @@ const AnalystComments = ({
...
-            '*': ['style'],
+            a: ['href', 'target', 'rel'],
           },
-        }));
+          transformTags: {
+            a: (_tagName, attribs) => ({
+              tagName: 'a',
+              attribs: {
+                href: attribs.href,
+                target: '_blankk',
```
elewis2, line 113: I realise this was there already, but is it definitely not a typo?

kttkjl: ` interface Tag { tagName: string; attribs: Attributes; text?: string | undefined; }`

is a legit parameter

app/views/MainView/index.tsx
kttkjl: Haven't tested it actually for when user does `anything` on the front-end

But as you can see, the async calls needs to happen, and if we do cleanup before the async calls goes through, there is still the overhead of attaching and detaching `eventlisteners` and cancelling calls more often than not

Of course, we can make the `updateToken` call to have a parameter (in seconds I believe) before the token expires to refresh 

But IMO, what we have currently is my suggestion on how this interaction would make most sense

app/services/management/auth.js
kttkjl: this doesn't need to happen, since for logout to work, the `keycloak` object instance should already be initialized

Tested with this line gone, nothing out of the ordinary happened 

THOUGH, this can come back to bite me for some obscure situation

elewis2 (approved):

app/views/ReportView/components/AnalystComments/index.tsx
kttkjl: oh sorry you meant the blankk

kttkjl: that's probably a typo then

elewis2 (approved):

PR #350, opened 2023-09-13
Chore/npm package updates

kttkjl:
[No ticket] NPM package updates -- it's about time

- Updated most packages 
- Fixed tests and settings files related to package updates
- React is staying at v17 for now, seems like it breaks more tests
- Let's not talk about how the js-keycloak adapter skipped 12 versions


.github/workflows/npm-test.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [12, 14, 16]
+        node: [16]
```
elewis2, line 10: add node 18, add node 20 if possible

elewis2 (changes requested):

.github/workflows/npm-test.yml
kttkjl: done, seem to work fine in the builds

elewis2 (approved):

PR #364, opened 2023-10-16
[DEVSU-2094] remove displayName column for rapid report

kttkjl:


app/views/ReportView/components/RapidSummary/columnDefs.ts
```diff
@@ -181,6 +175,13 @@ const sampleColumnDefs = [
     field: 'displayName',
     hide: false,
   },
+  {
```
elewis2, line 178: The genomicEvents column shouldn't be needed for the sample-info table (it's not a table of variants, it's just general info about the sample)

elewis2 (changes requested):

app/views/ReportView/components/RapidSummary/columnDefs.ts
kttkjl: Ah yes, bad commit by me -- Genomic Events wasn't even supposed to be here

kttkjl: should be fixed on commit [a533add](https://github.com/bcgsc/pori_ipr_client/pull/364/commits/a533add4b8847eae5bf18b39a8098b13a1c05a1d)

elewis2 (approved):

PR #369, opened 2023-10-25
[DEVSU-2086] DataTable text selectable and copiable

kttkjl:
Fixed some old code typing as well


app/components/DataTable/index.tsx
```diff
@@ -551,6 +555,7 @@ const DataTable = ({
               getRowNodeId={(data) => data.ident}
               onRowDragEnd={canReorder ? onRowDragEnd : null}
               editType="fullRow"
+              enableCellTextSelection
```
kttkjl, line 558: This is the only line that's relevant to the feature of this branch, everything else is developer debt

elewis2 (approved):

PR #376, opened 2023-10-31
[DEVSU-2078] Tumour summary edit for Rapid report

kttkjl:
- housekeeping of some types
  - conditional rendering of TumourSummaryEdit for different report types
  - renders form in sections without interdependency


elewis2 (changes requested):
I get an error when viewing this report:
report/c623d2c4-e947-4ffa-9ea6-092b7d404c3a/summary

kttkjl:
> I get an error when viewing this report: report/c623d2c4-e947-4ffa-9ea6-092b7d404c3a/summary

this is caused by KbMatches in Variants that are not from IPR-A and IPR-B being populated, the front-end assumes the backend data is correct in that there is only IPR-A and B, but this is fixed in PR #385

elewis2 (approved):

PR #378, opened 2023-11-02
[DEVSU-2113] add Zscore to CNVs, remove defaultly shown kIQR

kttkjl:


app/views/ReportView/components/CopyNumber/index.tsx
```diff
@@ -86,7 +86,7 @@ const CopyNumber = ({
             for (const {
               gene: {
                 expressionVariants: {
-                  tpm, rpkm, primarySiteFoldChange, diseasekIQR,
+                  tpm, rpkm, primarySiteFoldChange,
```
elewis2, line 89: in the other tables did diseaseZScore get added here or no?

elewis2 (changes requested):

app/views/ReportView/components/CopyNumber/index.tsx
kttkjl: This is old code that forced display of diseasKIQR because it was by default not shown, so removing this ends up not showing kIQR by default

elewis2 (approved):

PR #382, opened 2023-11-03
[DEVSU-2100] germline report linkout cell renderer

kttkjl:


app/views/GermlineView/components/LinkoutCellRenderer/index.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { ICellRendererParams } from '@ag-grid-community/core';
+
+const LINK_KEY_TO_URI = {
+  clinvar: 'https://www.ncbi.nlm.nih.gov/clinvar/variation/',
+  dbSnp: 'https://www.ncbi.nlm.nih.gov/snp/',
```
elewis2, line 7: I'm wondering if it makes sense for these base urls to be configurable or maybe storable in the db. Do we store other urls like this in the client code?

elewis2 (changes requested):

app/views/GermlineView/components/LinkoutCellRenderer/index.tsx
kttkjl: we store the graphkb base url in the build settings of this front-end code, but only because we control graphKB, the other linkouts are hardcoded in place like so

```
if (firstVal.match(/^\d+$/)) {
  link = `https://ncbi.nlm.nih.gov/pubmed/${firstVal}`;
  validLink = true;
} else if (urlRegex.test(firstVal)) {
  validLink = true;
}
```

the reason why we have this const is because I want this component to be reusable, and somewhere the base url can be edited and changes reflected on everything that uses this component.


>I'm wondering if it makes sense for these base urls to be configurable or maybe storable in the db. 

we're just storing it in another spot, the base URLs could theoretically be stored anywhere, just we'd have to know where to grab it

- If we store in the backend, if the base needs updating, then the backend needs to update it
- same thing with db
- The first two options is the same as what we're doing right now, but if we add an extra step in grabbing said base URL from the backend and THEN build it on the frontend, it's 1 extra step

Personally, I'd prefer this on the front-end, front-end issues should be fixable by the front-end

elewis2: all right, sounds reasonable!

elewis2 (approved):

PR #415, opened 2024-02-06
Feat/devsu 2201 sv burden editable

bnguyen-bcgsc:
- DEVSU-2201
- Add editable fields (sv count, sv percentile) to Tumour Summary Edit for both Genomic and Rapid reports
- Hardcode primary as the default comparator analysis role for new mutation burden data in Rapid reports
- Flag existing errors in Genomic report view and Rapid report view (code as TODO)


app/views/ReportView/components/RapidSummary/index.tsx
```diff
@@ -193,10 +193,9 @@ const RapidSummary = ({
             const burdenResp = await api.get(`/reports/${report.ident}/mutation-burden`).request();
             if (burdenResp[0].qualitySvCount == null) {
               setPrimaryBurden(null);
-              console.log(burdenResp[0]);
             } else {
               setPrimaryBurden(burdenResp[0]);
-              console.log(burdenResp[0]);
+              console.log(burdenResp[0]); // For debug purpose
```
kttkjl, line 198: should erase this from actual production code, probably forgot about this?

app/components/TumourSummaryEdit/index.tsx
```diff
@@ -150,6 +148,7 @@ const TumourSummaryEdit = ({
         if (mutationBurden?.ident) {
           apiCalls.push(api.put(`/reports/${report.ident}/mutation-burden/${mutationBurden.ident}`, newMutationBurdenData, {}));
         } else {
+          newMutationBurdenData.role = 'primary'; // Default role of new mutation burden data for reports with no existing analysis per ClinInfo team
```
kttkjl, line 151: This works, but that's a nono in React, if you need to change an attribute of an object in a state, you should use destructuring -- but even in this case 
`newMutationBurdenData.role = { ...newMutationBurdenData, role: 'primary' }` is unacceptable because it violates React principles where you should use the set<x> function (setNewMutationBurden) to change the state

instead of 
```
apiCalls.push(api.post(`/reports/${report.ident}/mutation-burden`, newMutationBurdenData, {}));
```

do
```
apiCalls.push(api.post(`/reports/${report.ident}/mutation-burden`, { ...newMutationBurdenData, role: 'primary' }, {}));
```

kttkjl (changes requested):
There are some changes here that coincide with PR#411, and some that will most definitely conflict with this PR, will pull #411 in first -- sorry about the confusion @bnguyen-bcgsc, that PR was slated as a hotfix but didn't get in on time

app/views/ReportView/components/GenomicSummary/index.tsx
```diff
@@ -197,6 +197,7 @@ const GenomicSummary = ({
           setVariants(sorted);
           setVariantCounts(counts);
           if (loadedDispatch) {
+            // TODO
```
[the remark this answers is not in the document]
bnguyen-bcgsc: yes my intellisense is throwing 'type {} has no call signature' for loadedDispatch. Maybe this is working as intended already and it's just intelli-nonsense but I flagged just in case

app/views/ReportView/components/RapidSummary/index.tsx
bnguyen-bcgsc: that was my bad I will remove this right away

app/components/TumourSummaryEdit/index.tsx
bnguyen-bcgsc: This is the elegant solution I was looking for. I'll make the proposed change now, thanks Jacky!

app/views/ReportView/components/RapidSummary/index.tsx
```diff
@@ -188,6 +189,18 @@ const RapidSummary = ({
...
+          try {
+            const burdenResp = await api.get(`/reports/${report.ident}/mutation-burden`).request();
+            if (burdenResp[0].qualitySvCount == null) {
+              setPrimaryBurden(null);
+            } else {
+              setPrimaryBurden(burdenResp[0]);
+            }
+          } catch (e) {
+            // mutation burden does not exist in records before this implementation, and no backfill will be done on the backend, silent fail this
+            console.error('mutation-burden call error', e?.message);
```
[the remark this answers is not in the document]
bnguyen-bcgsc: Hi Abbey, yes it is indeed a 404. I followed the implemented code in the Genomic Report View (line 155) for this section. I'm not sure if there's a way to check for the data's availability from the front end without calling the API

PR #416, opened 2024-02-07
[DEVSU-2148] tumour summary fixes

kttkjl:
Not EXACTLY related to DEVSU-2148, but while working on that ticket, the bugs were discovered, this is originally PR #411

Bugs fixed: 
- ApiCallSet settings were incorrect, logic also did not make sense and would cause problems, specifically TumourSummaryEdit/index.tsx line 167-173 
- mutationBurden creation is defaulted to role = primary
- Added default field values to tumourSummaryEdit instead of checking for an object

### DIFF
--- a/app/components/TumourSummaryEdit/index.tsx
+++ b/app/components/TumourSummaryEdit/index.tsx
@@ -353,34 +353,29 @@ const TumourSummaryEdit = ({
   );
 
   const tmburMutBurSection = useMemo(() => {
-    if (reportType === 'rapid') {
-      return (
-        <>
-          <TextField
-            className="tumour-dialog__text-field"
-            label="genomeSnvTmb"
-            value={newTmburMutData?.genomeSnvTmb ?? 0}
-            name="genomeSnvTmb"
-            onChange={handleTmburChange}
-            variant="outlined"
-            fullWidth
-            type="number"
-          />
-          <TextField
-            className="tumour-dialog__text-field"
-            label="genomeIndelTmb"
-            value={newTmburMutData?.genomeIndelTmb ?? 0}
-            name="genomeIndelTmb"
-            onChange={handleTmburChange}
-            variant="outlined"
-            fullWidth
-            type="number"
-          />
-        </>
-      );
-    }
-    return null;
-  }, [reportType, newTmburMutData?.genomeSnvTmb, newTmburMutData?.genomeIndelTmb, handleTmburChange]);
+    <>
+      <TextField
+        className="tumour-dialog__text-field"
+        label="genomeSnvTmb"
+        value={newTmburMutData?.genomeSnvTmb ?? 0}
+        name="genomeSnvTmb"
+        onChange={handleTmburChange}
+        variant="outlined"
+        fullWidth
+        type="number"
+      />
+      <TextField
+        className="tumour-dialog__text-field"
+        label="genomeIndelTmb"
+        value={newTmburMutData?.genomeIndelTmb ?? 0}
+        name="genomeIndelTmb"
+        onChange={handleTmburChange}
+        variant="outlined"
+        fullWidth
+        type="number"
+      />
+    </>;
+  }, [newTmburMutData?.genomeSnvTmb, newTmburMutData?.genomeIndelTmb, handleTmburChange]);
 
   return (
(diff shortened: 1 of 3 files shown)


app/components/TumourSummaryEdit/index.tsx
```diff
@@ -323,32 +339,27 @@ const TumourSummaryEdit = ({
...
-          onChange={handleMutationBurdenChange}
-          variant="outlined"
-          fullWidth
-          type="number"
-        />
-      );
-    }
-    return null;
-  }, [newMutationBurdenData, handleMutationBurdenChange]);
+  const mutBurDataSection = (
```
bnguyen-bcgsc, line 342: Hi Jacky, is there a reason why use don't use the useMemo() hook anymore for this section specifically? Thanks!

app/components/TumourSummaryEdit/index.tsx
```diff
@@ -323,32 +339,27 @@ const TumourSummaryEdit = ({
...
+      onChange={handleMutationBurdenChange}
+      variant="outlined"
+      fullWidth
+      type="number"
+    />
+  );
 
   const tmburMutBurSection = useMemo(() => {
-    if (newTmburMutData) {
+    if (reportType === 'rapid') {
```
bnguyen-bcgsc, line 356: Does this mean that these fields will not be editable for Genomic reports anymore?

app/components/TumourSummaryEdit/index.tsx
```diff
@@ -323,32 +339,27 @@ const TumourSummaryEdit = ({
...
-        />
-      );
-    }
-    return null;
-  }, [newMutationBurdenData, handleMutationBurdenChange]);
+  const mutBurDataSection = (
+    <TextField
+      className="tumour-dialog__text-field"
+      label="Mutation Burden (Mut/Mb)"
+      value={newMutationBurdenData?.totalMutationsPerMb ?? 0}
```
bnguyen-bcgsc, line 346: Should we change the default value of the field from '0' to 'null'? I know it won't save without a specific event that triggers the onChange but just incase the ClinInfo analysts see the value of 0 and wonder if it's 0 in the DB

app/components/TumourSummaryEdit/index.tsx
kttkjl: I believe your open PR takes care of this, which should remove this check

app/components/TumourSummaryEdit/index.tsx
kttkjl: That's a good question actually, I'm not too sure about the numbers but I removed useMemo here because the rendering of this one particular component doesn't seem worth enough (in terms of recalculation efficiency) to put into a memo hook

app/components/TumourSummaryEdit/index.tsx
kttkjl: Good question actually

Originally, this was made into a 0 (anything non null) because of MUI yelling at us
image
And because explanations [here](https://react.dev/reference/react-dom/components/input#im-getting-an-error-a-component-is-changing-an-uncontrolled-input-to-be-controlled)

We do have null values in the DB though so I understand where you are coming from -- currently changing from either 0 or '' into null doesn't seem to affect how the API calls are being made so I'd go with default values here

bnguyen-bcgsc: Ahh I understand now. Sounds great, I will change the default values on my end to keep it consistent with yours :)

elewis2: > changing from either 0 or '' into null doesn't seem to affect how the API calls are being made
Is the API currently being sent 0 either way?

elewis2 (approved):

app/components/TumourSummaryEdit/index.tsx
kttkjl: @elewis2 that was the question I was asking myself too when Bill posted the question, fortunately, we won't be sending the 0 if the field isn't touched -- the `isXDirty` field takes care of that, so that default 0 isn't being sent if the user doesn't specifically put 0 

Though that has the situation where the user wants to send a 0, but if the field is defaulted to 0, they won't see the 0 being updated to the backend

PR #417, opened 2024-02-08
[DEVSU-2179] report specific appendix

kttkjl:


app/context/ResourceContext/index.tsx
```diff
@@ -13,42 +13,57 @@ const REPORTS_BLOCK = [];
...
         setReportSettingAccess(true);
+        setReportEditAccess(true);
+      }
+    }
+  }, [groups]);
+
+  /**
+   * Check report specific permissions if user isn't admin
+   */
+  useEffect(() => {
```
kttkjl, line 51: @elewis2 If there are any other details you would like to add here -- currently they can edit if they are added to the report, did we want to more specifically check the permissions?

elewis2: No, I think this makes sense for now.

elewis2 (approved):

PR #418, opened 2024-02-13
Feat/DEVSU-2213-cd8t-score-editable

bnguyen-bcgsc:
- DEVSU-2213
- Add CD8 T cell score and percentile editable fields to genomic reports


app/components/TumourSummaryEdit/index.tsx
```diff
@@ -323,52 +378,95 @@ const TumourSummaryEdit = ({
     return null;
   }, [handleClicked, handleDelete, handleKeyDown, newMicrobialData]);
 
+  const tCellCd8DataSection = useMemo(() => {
+    return (
+      <>
+        <TextField
+          className="tumour-dialog__text-field"
+          label="CD8+ T Cell Score"
+          value={newTCellCd8Data?.score ?? 0}
```
elewis2, line 387: when the value is null, can we display no value here instead of 0? (here and in similar cases in tumour summary edit)

elewis2 (changes requested):

app/components/TumourSummaryEdit/index.tsx
kttkjl: react yells at you, this is a running problem, if we do want it to be null (no value) then the console of the browser will be spammed

elewis2: what's happening in the display when you delete the 0?

kttkjl: empty string 
image

elewis2 (approved):

elewis2 (approved):

PR #444, opened 2024-03-25
DEVSU-2246 Report Counts Missing in Projects

bnguyen-bcgsc:
- Fix issue where number of reports was not showing in projects
- Implement admin rights to Number of Reports and Number of User columns
- General linting

### DIFF
--- a/app/views/ProjectsView/columnDefs.ts
+++ b/app/views/ProjectsView/columnDefs.ts
@@ -16,7 +16,7 @@ const readOnlyColDefs = [
   },
   {
     headerName: 'Number of reports',
-    valueGetter: ({ data }) => data?.reports?.length,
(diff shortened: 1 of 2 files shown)


app/views/ProjectsView/columnDefs.ts
```diff
@@ -16,7 +16,7 @@ const readOnlyColDefs = [
   },
   {
     headerName: 'Number of reports',
-    valueGetter: ({ data }) => data?.reports?.length,
+    valueGetter: ({ data }) => Number(data?.reportCount),
```
kttkjl, line 19: not sure why the API comes back as string representation, but I'm not going to question it

kttkjl (approved):

PR #445, opened 2024-03-26
DEVSU-2193 Print Report Styling

bnguyen-bcgsc:
- Reformat key alteration categories in print table
- Change 'captiv 8' spelling to 'CAPTIV-8'
- Realign logo with other components in beta printview
- Increase report title font size in beta printview
- Add separating commas between key alterations variants


elewis2 (changes requested):
remove final comments in variant lists in key alterations table

app/components/SummaryPrintTable/index.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-nested-ternary */
```
kttkjl, line 1: you know the rules and so do i

elewis2 (approved):

app/components/SummaryPrintTable/index.tsx
```diff
@@ -12,6 +12,15 @@ type SummaryPrintTableProps = {
   renderValue?: (value: any) => unknown;
 };
 
+const variantTypes = ['cnv', 'smallMutation', 'structuralVariant', 'expression'];
+
+const variantToStrings: Dictionary<string> = {
```
kttkjl, line 17: weird, was the import from lodash needed? I thought the type inferrence by typescript was enough

kttkjl (approved):
Code looks good, but not sure about the import type from lodash

app/components/SummaryPrintTable/index.tsx
bnguyen-bcgsc: oh yea I tried the native typescript dict but it did not let me index with strings so I had to use the lodash dict

PR #453, opened 2024-04-15
feat/DEVSU-2256-add-toggle-for-allowNotifications

Nithriel:
Features:
- Add allowNotifications toggle


bnguyen-bcgsc (approved):

app/components/NavBar/components/UserSettingsDialog/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+    }
+  }, [checkboxSelected, editData, isSigned, showConfirmDialog, onClose, showErrorSnackbar]);
+
+  return (
+    <Dialog open={isOpen} maxWidth="sm" fullWidth className="edit-dialog">
+      <DialogTitle>Edit User Settings</DialogTitle>
+      <DialogContent>
+        <div>
+          <FormControlLabel
+            label="Allow email notifications?"
```
elewis2, line 86: can we change this to 'Allow email notifications to be sent to <user's email>', with no question mark (specifiying the email in case users wonder if the notifications they're permitting are for or about them), and put the check box on the right hand side of the text?

elewis2:
I was able to toggle my notifs off but not back on again.

elewis2 (changes requested):

app/components/NavBar/components/UserSettingsDialog/index.tsx
elewis2: Noting here that this was discussed in another forum and we decided on left hand side check box

bnguyen-bcgsc (approved):

elewis2 (approved):

PR #464, opened 2024-05-07
Feat/devsu 2275 make summary type configurable

elewis2:
Makes summary type selectable when creating a new report. This allows users to create new reports using eg the rapid report summary instead of the genomic report summary.

### DIFF
--- a/app/views/ReportView/components/Summary/index.tsx
+++ b/app/views/ReportView/components/Summary/index.tsx
@@ -11,19 +11,20 @@ const Summary = ({
   templateName,
   ...props
 }: SummaryProps): JSX.Element => {
-  if (templateName === 'probe') {
+  const summarySection = props.visibleSections.find(element => ['summary-genomic', 'summary-tgr', 'summary-pcp', 'summary-probe'].includes(element));
+  if (summarySection === 'summary-probe') {
     return (
       <ProbeSummary {...props} />
     );
   }
 
-  if (templateName === 'pharmacogenomic') {
+  if (summarySection === 'summary-pcp') {
     return (
       <PharmacoGenomicSummary {...props} />
     );
   }
 
-  if (templateName === 'rapid') {
+  if (summarySection === 'summary-tgr') {
     return (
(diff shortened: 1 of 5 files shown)


app/views/ReportView/components/Summary/index.tsx
```diff
@@ -11,36 +11,43 @@ const Summary = ({
   templateName,
   ...props
 }: SummaryProps): JSX.Element => {
-  if (templateName === 'probe') {
+  // TODO remove backup template name checks when data is updateed in prod
+  const summarySection = props.visibleSections.find(element => ['summary-genomic', 'summary-tgr', 'summary-pcp', 'summary-probe'].includes(element));
```
bnguyen-bcgsc, line 15: I got an error with this line when trying to print the report: "Cannot read properties of undefined (reading 'find').
TypeError: Cannot read properties of undefined (reading 'find')"

app/views/ReportView/components/Summary/index.tsx
```diff
@@ -11,36 +11,43 @@ const Summary = ({
...
+  if (summarySection === 'summary-probe' || templateName === 'probe') {
     return (
       <ProbeSummary {...props} />
     );
   }
 
-  if (templateName === 'pharmacogenomic') {
+  if (summarySection === 'summary-pcp' || templateName === 'pharmacogenomic') {
+    console.log(templateName);
```
bnguyen-bcgsc, line 23: Are we keeping the console logs on purpose?

bnguyen-bcgsc (changes requested):

app/views/ReportView/components/Summary/index.tsx
elewis2: ah, no

app/views/ReportView/components/Summary/index.tsx
elewis2: fixed, had to update the SummaryProps type

bnguyen-bcgsc (approved):

PR #481, opened 2024-05-22
Feat/devsu 2271 enable condensed view for nongenomic reports

elewis2:
adds condensed view option to nongenomic report types; combines patient and tumour info in condensed view for rapid.


bnguyen-bcgsc (changes requested):
I made some small changes to the code and also, are we allowing Rapid reports to have both standard layout and condensed layout? Or are we making the condensed layout the default print layout for it?

bnguyen-bcgsc (approved):

PR #496, opened 2024-06-07
Feat/DEVSU-2333 move biopsy collection date up to the patient summary

sshugsc:
[DEVSU-2333](https://www.bcgsc.ca/jira/browse/DEVSU-2333)
- add biopsy collection date to patient information


app/views/ReportView/components/PatientInformation/index.tsx
```diff
@@ -40,6 +41,11 @@ const PatientInformation = ({
   useEffect(() => {
     if (report?.ident) {
       const getData = async () => {
+        const appendicesResp = await api.get(`/reports/${report.ident}/appendices`).request();
+        let biopsyCollectionDate;
+        if (appendicesResp.sampleInfo) {
+          biopsyCollectionDate = appendicesResp.sampleInfo[0]['Collection Date'];
```
kttkjl, line 47: is the API setup so that the first sampleInfo is always going to be what you're looking for? or is this a trival thing

bnguyen-bcgsc (dismissed):

app/views/ReportView/components/PatientInformation/index.tsx
sshugsc: Good question! Sample Info sometimes has both "Tumor" and "Constitutional" sample collection date ("Tumor" always the first one), but sometimes only has "Constitutional" sample collection date. Will double check with Eleanor, when she's back.

elewis2 (approved):

app/views/ReportView/components/PatientInformation/index.tsx
elewis2: I've pinged clininfo on DEVSU-2333 to confirm but I think it'll be the tumour sample.

elewis2: Ok, yes it'll be the tumour sample collection date. If we don't have that, leave the field blank.

elewis2: Is there a way to check whether this is the tumour sample or not?

elewis2 (changes requested):

app/views/ReportView/components/PatientInformation/index.tsx
```diff
@@ -44,8 +44,11 @@ const PatientInformation = ({
     if (report?.ident) {
       const getData = async () => {
         let biopsyCollectionDate;
-        if (appendices && appendices.sampleInfo && appendices.sampleInfo[0]['Sample'] === 'Tumour') {
-          biopsyCollectionDate = appendices.sampleInfo[0]['Collection Date'];
+        if (appendices && appendices.sampleInfo) {
+          appendices.sampleInfo.map((info)=>{
```
kttkjl, line 48: instead of map you can just use forEach, because map creates another dynamic array -- takes up a bit more space than searching an already existing array

bnguyen-bcgsc: Can she use .find() too?

kttkjl: `find` works fine too yeah, it short circuits which is even better

kttkjl (approved):

bnguyen-bcgsc (approved):

app/views/ReportView/components/PatientInformation/index.tsx
sshugsc: learned!

elewis2 (approved):

PR #497, opened 2024-06-10
Feat/devsu 2336 add frontend for setting displaying variant text

sshugsc:
[DEVSU-2336](https://www.bcgsc.ca/jira/browse/DEVSU-2336)
- add variant-text to admin view


bnguyen-bcgsc (dismissed):

app/views/AdminView/components/VariantText/components/AddEditVariantText/index.scss
```diff
@@ -0,0 +1,50 @@
+@import '@/styles/_theme.module.scss';
+
+.edit-dialog {
```
kttkjl, line 3: I'd call this `.variant-text__edit-dialog` and put everything under, like so,

```
.variant-text__edit-dialog {
    &__section-title {
        margin-top: 24px;
        margin-bottom: 24px;
    }

    .add-item {
        &__form-container {
            flex-direction: row;
        }

        &__select-field {
            margin: 8px;
        }
    }

    .text-field {
        margin: 8px;
    }

    .IPRWYSIWYGEditor {
        &__content {
            margin: 8px;
            padding: 0.5rem;
            border: 1px solid $palette__grey--300;
            min-height: 250px;
        }

        &__toolbar {
            z-index: +1;
            position: sticky;
            top: 0;
            padding: 8px;
            width: 100%;
            background-color: $palette__common--white;
            border-bottom: none;
        }
    }
}
```

due to how specificity goes, changing `.IPRWYSIWYEditor` under no other classes will cause unexpected changes in the future

app/views/AdminView/components/VariantText/components/AddEditVariantText/index.scss
```diff
@@ -0,0 +1,50 @@
...
+        padding: 8px;
+        width: 100%;
+        background-color: $palette__common--white;
+        border-bottom: none;
+    }
+
+  }
+
+#template-select-label {
+    margin: 8px;
```
kttkjl, line 45: MUI has something called `sx` that you can use on the component itself to style something instead of using an ID, using an ID is only when you have no other choices

```
          <Select
            defaultValue=""
            id="projects-select"
            label="Project"
            variant="outlined"
            className="add-item__select-field"
            fullWidth
            required
            sx={{ margin: '8px' }}
            onChange={handleProjectChange}
          >
```

app/views/AdminView/components/VariantText/columnDefs.ts
```diff
@@ -0,0 +1,49 @@
...
+    minWidth: 90,
+  },
+  {
+    headerName: 'Cancer Type',
+    valueGetter: ({ data }) => data.cancerType,
+    minWidth: 90,
+  },
+  {
+    headerName: 'Text',
+    cellRenderer: ({ data }) => (data.text ? `${data.text.substring(0, 300)}....` || '' : ''),
```
kttkjl, line 35: I'm not sure about this, due to having the text `....` always being there if string length less <300, I can understand where the motivation comes from, but I don't think it is working as intended

the `....` is there for when the text is overflown and blocked off, but the way the table is rendered makes it so that the whole text cell is rendered and brought to size, making this .... an artefact

had the cell been a 1 row height instead, the user can click this cell and expand the text, but in this case the text is already expanded so I think we can leave the .... out unless you decide to make this fancy and expand out

kttkjl (changes requested):
Looks great mostly, but some finesse in css and styles needed for possible conflicts with other styles

app/views/AdminView/components/VariantText/components/AddEditVariantText/index.scss
```diff
@@ -0,0 +1,50 @@
+@import '@/styles/_theme.module.scss';
+
+.edit-dialog {
+    &__section-title {
```
kttkjl, line 4: actually, is this even used?

app/views/AdminView/components/VariantText/components/AddEditVariantText/index.tsx
```diff
@@ -0,0 +1,263 @@
...
+import './index.scss';
+import { MenuBar } from '@/components/IPRWYSIWYGEditor';
+import {
+  useEditor, EditorContent,
+} from '@tiptap/react';
+import StarterKit from '@tiptap/starter-kit';
+import Underline from '@tiptap/extension-underline';
+import sanitizeHtml from 'sanitize-html';
+import { VariantTextType, TemplateType } from '@/common';
+import { ProjectType } from '../../../../types';
```
kttkjl, line 30: this can and should be imported from @common

elewis2 (changes requested):
looks good. one issue: when you create variant text, then try to create another variant text, the text from the old variant is copied over into the new variant text field. something probably needs to get reset to null when the window is closed

elewis2 (changes requested):
Is it possible with the current api configuration to make the project and template optional?

bnguyen-bcgsc:
@elewis2 right now the edit function only allows for editing the text. Do you think we should allow users to edit the other fields too? For example like changing the variant or adding template and project afterwards since they are optional now

app/views/AdminView/components/VariantText/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+    setEditingData(rowData);
+    setIsEditing(true);
+  }, []);
+
+  const handleOnAdd = useCallback(() => {
+    setIsAdding(true);
+  }, []);
+
+  const handleOnDelete = useCallback(async (rowData: VariantTextType) => {
+    try {
```
kttkjl, line 61: Visually, looks good, there is one interaction missing though

When a user deletes a record on VariantText main view, the table doesn't get updated 

use the [Appendix onDelete as reference](https://github.com/bcgsc/pori_ipr_client/blob/0e8d55d7c37515a3f75cac26c6ca057b8cd33557/app/views/AdminView/components/Appendices/index.tsx#L60) to update the `variantText` state

kttkjl (changes requested):
Visually looks good, just need that one interaction fixed for up to date data

app/views/AdminView/components/VariantText/index.tsx
sshugsc: just pushed

kttkjl: rats, too slow

kttkjl (approved):

elewis2:
> @elewis2 right now the edit function only allows for editing the text. Do you think we should allow users to edit the other fields too? For example like changing the variant or adding template and project afterwards since they are optional now
Possibly yes but this could go wrong - the user would potentially be overwriting another record but not editing that other record directly, if that makes sense? Let's address this in a separate ticket

bnguyen-bcgsc (approved):

elewis2 (approved):

PR #511, opened 2024-07-04
[DEVSU-2265] dynamic split table

kttkjl:
PrintTables now take first row of rendered table to determine widths for all tds 

Last row of cut-off table doesn't get thrown out

Fixed some styling here and there

There is a known bug where sometimes a auto-row-spanned row is split into two 

Reports to test on:
Genomic: 1461, 1797


bnguyen-bcgsc (changes requested):
Left comment on jira ticket

bnguyen-bcgsc (approved):

PR #522, opened 2024-07-17
[DEVSU-2377] tumour summary edit no props

kttkjl:
Merge this after 2362 is merged, this is dependent on that branch, and that bugfix is dependent on this to work properly as well


kttkjl:
Closing PR, merged into 2362, will fix both problems at once due to how interconnected it is

PR #523, opened 2024-07-17
[DEVSU-2362, 2377, 2396] TumourSummary bugfixes

kttkjl:


kttkjl:
As discussed on RC, logic is as follows

```
if there's an empty list, 'Not detected'
if there's a non-empty list that only contains 'None', display 'Not detected'
if there's a non-emtpy list that contains non-'None' values and 'None', display the non-'None' values but not the 'None'
```

PR #534, opened 2024-07-22
feat/DEVSU-2389 update groups detailed view

bnguyen-bcgsc:
- DEVSU-2389
- Update detailed view for Groups to display users by their full name instead of object id

### DIFF
--- a/app/components/DataTable/components/DetailDialog/index.tsx
+++ b/app/components/DataTable/components/DetailDialog/index.tsx
@@ -48,12 +48,16 @@ const PrimitiveAttribute = ({ keyString, value }: PrimitiveAttributeProps) => (
 
 type ObjectAttributesProps = {
   obj: object;
-  columnMapping: Record<string, string>
+  objKey?: string;
+  columnMapping: Record<string, string>;
+  arrayKeyGetter?: Record<string, (val) => string>;
 };
 
 const ObjectAttributes = ({
   obj,
+  objKey,
   columnMapping,
+  arrayKeyGetter,
 }: ObjectAttributesProps) => {
   if (obj === null || obj === undefined) {
     return null;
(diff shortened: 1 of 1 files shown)


app/components/DataTable/components/DetailDialog/index.tsx
```diff
@@ -89,7 +89,7 @@ const ObjectAttributes = ({
               expandIcon={<ExpandMoreIcon />}
             >
               <Typography variant="subtitle2">
-                {`${idx}: ${mappedValIsArray ? '[' : '{'}`}
+                {`${!!mappedVal?.username ? `${mappedVal?.firstName} ${mappedVal?.lastName}` : idx}: ${mappedValIsArray ? '[' : '{'}`}
```
elewis2, line 92: I take it this isn't something we can customize for individual tables?

elewis2: Could we add a second default - if 'name' is present, then display name?

bnguyen-bcgsc: Changes pushed

app/components/DataTable/components/DetailDialog/index.tsx
```diff
@@ -89,7 +89,9 @@ const ObjectAttributes = ({
               expandIcon={<ExpandMoreIcon />}
             >
               <Typography variant="subtitle2">
-                {`${idx}: ${mappedValIsArray ? '[' : '{'}`}
+                {!!mappedVal?.name ? `${mappedVal.name}` : ''}
+                {!!mappedVal?.username ? `${mappedVal?.firstName} ${mappedVal?.lastName}` : ''}
+                {`${!mappedVal?.username && !mappedVal?.name ? idx : ''}: ${mappedValIsArray ? '[' : '{'}`}
```
kttkjl, line 94: this is supposed to be a generic component, adding all of these defeats the purpose

kttkjl (changes requested):

app/components/DataTable/components/DetailDialog/index.tsx
bnguyen-bcgsc: Can I make them optional props?

kttkjl (approved):

PR #536, opened 2024-07-25
remove column not wanted by clininfo

elewis2:
Remove column not wanted by CLININFO


bnguyen-bcgsc (changes requested):
I believe this column has to be removed in utils/reportsColumns.js as well

bnguyen-bcgsc (approved):

PR #546, opened 2024-07-31
bugfix/DEVSU-2410 Automatically assign signing users to report

bnguyen-bcgsc:
- DEVSU-2410
- Allow signing users with report assignment access but not admin and manager to sign report and be assigned dynamically to said report
- If the user gets removed from report, they cannot delete signatures

### DIFF
--- a/app/components/SignatureCard/index.tsx
+++ b/app/components/SignatureCard/index.tsx
@@ -1,5 +1,5 @@
 import React, {
-  useState, useEffect, useMemo,
+  useState, useEffect, useMemo, useCallback,
 } from 'react';
 import {
   Paper,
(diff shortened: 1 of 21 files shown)


elewis2 (dismissed):

app/components/SignatureCard/index.tsx
```diff
@@ -36,23 +39,41 @@ const SignatureCard = ({
...
       }
     }
-  }, [signatures, type]);
+  }, [signatures, type, setRole]);
 
-  const handleSign = () => {
-    onClick(true, type);
+  const handleSign = async () => {
+    try {
+      const newReport = await api.post(
```
kttkjl, line 66: separating the API calls on PUT requests to `report` and `signatures` on different components will cause discrepencies on the UI whenever a user interacts with "sign".

This is actually a good idea to put the API calls on the card itself, let's update the signatures here instead and send the callback for `onClick(true, type)` whenever both calls succeed, and we can also update the report here as well, instead of 
`const handleSign = useCallback(async (signed: boolean, role: SignatureUserType) ` we can have
`const handleSign = useCallback(async (signed: boolean, updatedSignatures)` instead

kttkjl (changes requested):

kttkjl (changes requested):
Sorry, found some bugs, addressed in https://github.com/bcgsc/pori_ipr_client/pull/553

kttkjl (approved):

PR #556, opened 2024-08-12
[DEVSU-2427] bugfix: add user dialog

kttkjl:
Add allProjectsAccess and make default value for project/groups to be null instead so the form sees empty array as "dirty"


elewis2 (changes requested):
handle case when user has no projects and also doesn't have all-projects-access role, without getting stuck at spinning icon

kttkjl:
> handle case when user has no projects and also doesn't have all-projects-access role, without getting stuck at spinning icon

[cb93716](https://github.com/bcgsc/pori_ipr_client/pull/556/commits/cb9371655e53d6432ebc24770ca8d7f5667c6fe1)

elewis2 (approved):

PR #559, opened 2024-08-14
Bugfix/devsu 2432 restrict all project access granting

elewis2:
Update DataTable to accept a function for canEdit. 

Update Groups view to not restrict group editing as follows:
 admin can edit all groups
admin or allProjectsAccess+manager is required to edit allProjectsAccess

Update Users view to restrict user editing as follows:
admin or allProjectsAccess+manager is required to grant allProjectsAccess membership

### DIFF
--- a/app/common.d.ts
+++ b/app/common.d.ts
@@ -56,7 +56,6 @@ type GroupType = {
   ident: string;
   name: string;
   users: UserGroupMemberType[];
-  owner: UserType;
   description: string;
 };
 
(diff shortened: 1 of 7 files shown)


kttkjl (changes requested):
I can't approve this, this is in violation of where we defined group-level permissions and moving all of that into individual components where this logic cannot be re-used

`checkGroupdEditPermissions` should be in a reusable context

elewis2:
> I can't approve this, this is in violation of where we defined group-level permissions and moving all of that into individual components where this logic cannot be re-used
> 
> `checkGroupdEditPermissions` should be in a reusable context

ok, moved checkGroupEditPermissions into the resource context (still needed a wrapper func to provide the group name from rowData)

PR #563, opened 2024-08-15
bugfix/DEVSU-2437 display signature section even if no comments

elewis2:
Displays the signatures boxes in the analyst comments section even if there is no analyst text yet.


kttkjl (changes requested):
`comments` in 151 needs a removal

unrelated to this PR: 
line 81 report needs to be removed as it is unnecessary too

else looks good

PR #566, opened 2024-09-12
Chore: DEVSU-2433 Realign checkbox to add user to graphkb in add edit user dialog

bnguyen-bcgsc:
- Realign checkbox to add user to graphkb in add edit user dialog

### DIFF
--- a/app/views/AdminView/components/Users/components/AddEditUserDialog/index.tsx
+++ b/app/views/AdminView/components/Users/components/AddEditUserDialog/index.tsx
@@ -127,6 +127,7 @@ const AddEditUserDialog = ({
         }
       } catch (projectGroupErr) {
         snackbar.error('Failed to retrieve list of project and groups for current user.');
+        // eslint-disable-next-line no-console
         console.error(projectGroupErr);
       } finally {
         setIsDataLoading(false);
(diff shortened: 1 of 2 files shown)


app/views/AdminView/components/Users/components/AddEditUserDialog/index.tsx
```diff
@@ -456,6 +457,7 @@ const AddEditUserDialog = ({
             <Checkbox checked={gkbAdd} onChange={handleCheckboxChange} />
           }
           disabled={Boolean(editData)}
+          sx={{ paddingLeft: '6px' }}
```
kttkjl, line 460: was adding this to add-user__checkbox not enough that warrants an in-line instead? 

If possible we'd like to do it in className instead since it's already setup, as future devs might just edit that one instead and this in-line here might cause hiccups

kttkjl (commented):
waiting on answer

app/views/AdminView/components/Users/components/AddEditUserDialog/index.tsx
bnguyen-bcgsc: yea you're right I'm changing it

kttkjl (approved):

PR #571, opened 2024-11-04
Feat/DEVSU-2445 update kbmatch multiple variants statements

bnguyen-bcgsc:
- DEVSU-2445 & DEVSU-2467
- Create kbMatches and kbMatchedStatement types
- Update KbMatches tables columns for arrays of variants
- Update KbMatchActionCellRenderer to comply with new data structure of kbMatchedStatements
- Update ArrayCellRenderer
- Update endpoints in KbMatches view
- Update coalesce middleware to comply with new data structure of kbMatchedStatements

*Test with DEVSU-2457 api build*

PR #572, opened 2024-11-06
bugfix/DEVSU-2445 rapid report kb matches bug fix

bnguyen-bcgsc:
*Extension of [https://github.com/bcgsc/pori_ipr_client/pull/571]*

- DEVSU 2445 & 2467 & 2505
- Update kb matches and evidence level splitting logic in rapid report with new kb matched statements response data structure
- Update valueGetter logic for observed variants column in kb matches tables
- Switch positions of observed variants and known variants columns back to original
- Rework kbMatches statement categories handling in smallMutations, copyNumber, structuralVariants, and expressionOutliers to work with new kbmatches data structure
- Fix ExpOutliersType missing in GeneType

### DIFF
--- a/app/views/ReportView/components/KbMatches/columnDefs.ts
+++ b/app/views/ReportView/components/KbMatches/columnDefs.ts
@@ -68,7 +68,13 @@ const columnDefs: ColDef[] = [
     valueGetter: (params) => {
       const { data: { kbMatches } } = params;
       if (kbMatches) {
-        const kbVariants = kbMatches?.map((match) => match.kbVariant).filter((kbVariant) => kbVariant !== undefined);
+        // const kbVariants = kbMatches?.map((match) => match.kbVariant).filter((kbVariant) => kbVariant !== undefined);
+        const kbVariants = kbMatches?.reduce((accumulator, match) => {
+          if (match.kbVariant !== undefined) {
+            accumulator.push(match.kbVariant);
+          }
+          return accumulator;
+        }, []);
         return kbVariants.join(', ');
       }
       return null;
@@ -100,7 +106,10 @@ const columnDefs: ColDef[] = [
             case ('mut'):
               variantArr.push(`${kbMatch?.variant.gene.name}:${kbMatch?.variant.proteinChange}`);
               break;
-            case ('msi' || 'tmb'):
+            case ('tmb'):
+              variantArr.push(kbMatch?.variant.kbCategory);
+              break;
+            case ('msi'):
               variantArr.push(kbMatch?.variant.kbCategory);
               break;
             default:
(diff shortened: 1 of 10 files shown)


app/views/ReportView/components/RapidSummary/index.tsx
```diff
@@ -57,7 +57,7 @@ const splitIprEvidenceLevels = (kbMatches: KbMatchType[]) => {
   return iprRelevanceDict;
 };
 
-const processPotentialClinicalAssociation = (variant: RapidVariantType) => Object.entries(getVariantRelevanceDict(variant.kbMatches))
+const processPotentialClinicalAssociation = (variant: any) => Object.entries(getVariantRelevanceDict(variant.kbMatches))
```
kttkjl, line 60: uhh ...?

bnguyen-bcgsc: srry typescript was throwing some errors I'll try to fix it

elewis2 (dismissed):

app/views/ReportView/components/KbMatches/columnDefs.ts
```diff
@@ -63,7 +63,21 @@ const columnDefs: ColDef[] = [
     sort: 'asc',
   },
   {
-    headerName: 'Observed Variant(s)',
+    headerName: 'Known Variants',
+    colId: 'kbVariant',
+    valueGetter: (params) => {
+      const { data: { kbMatches } } = params;
+      if (kbMatches) {
+        const kbVariants = kbMatches?.map((match) => match.kbVariant).filter((kbVariant) => kbVariant !== undefined);
```
kttkjl, line 71: we can probably just use `.filter` no?

`kbMatches?.filter((match) => match?.kbVariant);`

bnguyen-bcgsc: oh yea you're right let me simplify it

bnguyen-bcgsc: @kttkjl actually I tried it and it didn't work, we still have to extract kbVariant from each match with .map() right?

kttkjl: sorry, `match.kbVariant && match.kbVariant !== undefined`, maybe the shorthand is not enough

kttkjl: unless we want to keep `null, false, 0 and ''` values?

kttkjl: in which case we'll use reduce

```
kbMatches?.reduce((acc, match) => {
  if (match.kbVariant !== undefined) {
    acc.push(match.kbVariant);
  }
  return acc;
}, []);
```

kttkjl (approved):
Programmatically makes sense, just `reduce` vs `map` + `filter`

elewis2 (approved):

PR #584, opened 2025-01-17
bugfix/DEVSU-2549 cancerType not a valid property for variant texts

sshugsc:
[DEVSU-2549](https://www.bcgsc.ca/jira/browse/DEVSU-2549)
- fix bad request error when trying to edit variantText comment

### DIFF
--- a/app/views/AdminView/components/VariantText/index.tsx
+++ b/app/views/AdminView/components/VariantText/index.tsx
@@ -77,27 +77,11 @@ function VariantText(): JSX.Element {
     let cancelled = false;
     try {
       if (nextData) {
-        const sanitizedText = sanitizeHtml(nextData, {
-          allowedAttributes: {
-            a: ['href', 'target', 'rel'],
-          },
-          transformTags: {
-            a: (_tagName, attribs) => ({
-              tagName: 'a',
-              attribs: {
-                href: attribs.href,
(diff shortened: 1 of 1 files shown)


bnguyen-bcgsc (approved):

kttkjl (changes requested):
the onClose for this is a confirmation, why is this removed? What you're doing right here is just setting the front-end status of admin data, and no API request actually gets into the back end

sshugsc:
> the onClose for this is a confirmation, why is this removed? What you're doing right here is just setting the front-end status of admin data, and no API request actually gets into the back end

The API put request is at https://github.com/bcgsc/pori_ipr_client/blob/develop/app/views/AdminView/components/VariantText/index.tsx#L95 .
If we not remove that, it will do API put request twice.

kttkjl:
> > the onClose for this is a confirmation, why is this removed? What you're doing right here is just setting the front-end status of admin data, and no API request actually gets into the back end
> 
> The API put request is at https://github.com/bcgsc/pori_ipr_client/blob/develop/app/views/AdminView/components/VariantText/components/AddEditVariantText/index.tsx#L177 If we not remove that, it will do API put request twice.

Oh I see, the base component changed so the interaction's different

kttkjl (approved):

PR #589, opened 2025-02-12
[DEVSU-2557] kbmatch table re-assignment

kttkjl:
Multi-row select for tables of KbMatches view for Genomic 

Ctrl-click multi selection, or shift (but that does text selecting too)
 - Right click will bring out context menu for moving KbStatementIds to target table, if selected, prompts dialog
 - Menu beside table name will have extra option of toggling the dialog as well

Single selection:
- Right clicking a single row will bring out context menu to toggle dialog
- Selecting a row-specific action menu will also do the same

Known behaviors:
 - After moving statements, if there are any lingering entries (if you planned to move all of them), then it is IPR-API's duplicate entries with different idents, there is no way of selecting for priority either because the creation time is the same

### DIFF
--- a/app/components/DataTable/components/KbMatchesActionCellRenderer/index.tsx
+++ b/app/components/DataTable/components/KbMatchesActionCellRenderer/index.tsx
@@ -33,12 +33,14 @@ const KbMatchesActionCellRenderer = (props: ActionCellRendererProps) => {
   const { data } = props;
 
   const {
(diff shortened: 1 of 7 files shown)


kttkjl:
Broken tests, fixing

bnguyen-bcgsc (approved):

elewis2 (approved):

PR #599, opened 2025-02-25
[DEVSU-2574] genomic report pharmacogenomic table fix

kttkjl:
updated API call and value getters for updated schema


kttkjl:
I might have gotten this wrong on which API to change to, let me know

app/utils/kbMatchStatementsGeneValueGetter.ts
```diff
@@ -0,0 +1,51 @@
...
+    const kbMatchesNonNull = kbMatches?.filter((match) => !Array.isArray(match));
+    if (kbMatchesNonNull.length > 1) {
+      const geneName = [];
+      for (const kbMatch of kbMatchesNonNull) {
+        if (kbMatch?.variantType === 'msi') {
+          geneName.push('msi');
+        }
+        if (kbMatch?.variantType === 'tmb') {
+          geneName.push('tmb');
+        }
```
bnguyen-bcgsc, line 14: Should we add the 'sigv' check here too?

bnguyen-bcgsc (changes requested):

app/utils/kbMatchStatementsGeneValueGetter.ts
kttkjl: probably, that's the new thing that's been merged into develop eh?

elewis2: signature variant types won't have a gene

bnguyen-bcgsc (approved):

bnguyen-bcgsc (changes requested):
@kttkjl I believe the PharmacoGenomicSummary component also has to be updated (both columnDefs and index) since they still have some areas where the kbmatches are not updated to the new data structure

elewis2 (approved):

bnguyen-bcgsc (approved):

PR #604, opened 2025-03-04
[DEVSU-2587] fix pcp-tgr-kbmatches data update

kttkjl:
~~Should probably make `ArrayCell` more generic, but since this just a one off, maybe not~~

Just kidding, we did, `ArrayCellRenderer` now accepts complicated paths, and even accounts for arrays


kttkjl:
Actually hold off on this, there's more fields to fix asides from the specific example

bnguyen-bcgsc (approved):

kttkjl:
SHOULD be good to go, hoping tests runs

elewis2 (approved):

PR #607, opened 2025-03-11
DEVSU-2581 Update TemplateView to use react-query

OwenCoelho14:
DEVSU-2581 Update TemplateView to use react-query
- import and add QueryClientProvider to App.tsx
- include 'react-query' in package.json and package-lock.json
- change data fetching in TemplateView to use react query

### DIFF
--- a/app/views/TemplateView/index.tsx
+++ b/app/views/TemplateView/index.tsx
@@ -1,25 +1,40 @@
-import React, { useState, useEffect, useCallback } from 'react';
+import React, { useState, useCallback } from 'react';
 import { useSnackbar } from 'notistack';
+import { useQuery, useMutation, useQueryClient } from 'react-query';
 
 import DataTable from '@/components/DataTable';
 import api from '@/services/api';
 import AddEditTemplate from './components/AddEditTemplate';
 import columnDefs from './columnDefs';
 
+const fetchTemplates = async () => api.get('/templates', {}).request();
+
+const deleteTemplate = async (ident: string) => {
+  await api.del(`/templates/${ident}/signature-types`, {}, {}).request();
+  return api.del(`/templates/${ident}`, {}, {}).request();
+};
+
 const TemplateView = (): JSX.Element => {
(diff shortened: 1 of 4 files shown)


app/views/TemplateView/index.tsx
```diff
@@ -31,28 +46,19 @@ const TemplateView = (): JSX.Element => {
...
-      await api.del(`/templates/${rowData.ident}/signature-types`, {}, {}).request();
-      await api.del(`/templates/${rowData.ident}`, {}, {}).request();
-      setTemplates((prevVal) => prevVal.filter((template) => template.ident !== rowData.ident));
+      deleteMutation.mutate(rowData.ident);
       snackbar.enqueueSnackbar('Template deleted');
     } catch (err) {
       snackbar.enqueueSnackbar(`Error deleting template: ${err}`);
     }
-  }, [snackbar]);
+  }, [deleteMutation]);
```
kttkjl, line 61: missing snackbar here

kttkjl (changes requested):
Looks good, just missing dependency array param

Edit: and `snackbarUtils` change to use what we built on top of `notistack` instead of using `notistack` directly

app/views/TemplateView/index.tsx
```diff
@@ -1,25 +1,40 @@
...
+  const { data: templates = [] } = useQuery({
+    queryKey: ['templates'],
+    queryFn: fetchTemplates,
+  });
+
+  const deleteMutation = useMutation({
+    mutationFn: deleteTemplate,
+    onSuccess: () => {
+      queryClient.invalidateQueries({ queryKey: ['templates'] });
+      snackbar.enqueueSnackbar('Template deleted');
```
kttkjl, line 32: actually, we should use snackbar.success since we're here, 

`import snackbar from '@/services/SnackbarUtils';`

and then do `snackbar.success('Template deleted');`

same thing with 
`snackbar.error(`Error deleting template: ${err}`);`

kttkjl (approved):

PR #617, opened 2025-03-18
[DEVSU-2503] therapeutic options print ordering

kttkjl:
1. New `TherapeuticTargetPrintTable`
2. Move `PrintTable` styling out to be used with `TherapeuticTargetPrintTable`
3. Updated `!important` css attributes out of cut-off repeated table headers, now managed by pagedJs handlers
4. Cleanup on old components using old references


bnguyen-bcgsc (approved):

elewis2 (changes requested):
some genes getting lost in print display when the table is large, details on ticket

kttkjl:
Will need to put the coalescing logic into another pagedjs handler

elewis2 (changes requested):
when I test this I'm seeing some rows omitted and big gaps between adjacent coalesced rows. example on ticket

kttkjl:
Turns out the style fixing on `afterRendered` hook just won't work nicely (padding/margin messes up due to colSpan), forced to use `!important`, all other hooks doens't seem to work with varying reasons (either the DOM didn't render yet, so no elements are grabbed, or the page have not been split by pagedjs yet)

PR #619, opened 2025-03-19
[DEVSU-2468] update adminView/NewUserDialog to call to graphkb new user too

kttkjl:
Calls a new endpoint to create user on graphKb instead of putting it on a queue, not knowing when it finishes

Used in tandem with https://github.com/bcgsc/pori_ipr_api/pull/437


elewis2 (changes requested):
can the popups handle graphkb user creation failure separately from ipr user creation outcome?

kttkjl:
> can the popups handle graphkb user creation failure separately from ipr user creation outcome?

Good idea, updating

bnguyen-bcgsc (approved):

elewis2:
This doesn't appear to be showing any outcome for Graphkb user creation - ipr user successfully created, graphkb user requested but not successfully created, only popup displayed was for successful ipr user creation.

elewis2 (changes requested):
show result of graphkb user creation

kttkjl:
> show result of graphkb user creation

It >should< show creation result, tested it pretty thoroughly, what was the setup for your tests? @elewis2

elewis2 (approved):

PR #620, opened 2025-03-19
DEVSU-2603 Fix kbMatchedStatements coalesce logic

bnguyen-bcgsc:
- DEVSU-2603
- Fix coalesce issue where kbMatches arrays get nested in each other unintentionally when statements are coalesced. Now kbMatches records are concatenated in the same array, and only when they have different kbVariantId
- De-duplicate gene names, known variants, and observed variants in cases where there are multiple coalesced kbMatches
- Remove redundant bucket key creation logic

### DIFF
--- a/app/views/ReportView/components/KbMatches/coalesce.ts
+++ b/app/views/ReportView/components/KbMatches/coalesce.ts
@@ -101,8 +101,8 @@ const coalesceEntries = <T extends KbMatchedStatementType[]>(entries: T): Coales
         Object.entries(entry).forEach(([key, value]) => {
           if (Array.isArray(buckets[bucketKey][key])) {
             if (key === 'kbMatches' && Array.isArray(value)) {
-              const kbMatchesIdents = buckets[bucketKey][key].map((kbM) => kbM.kbVariantId).join();
-              const coalescedKbMatchesIdents = value.map((kbM) => kbM.kbVariantId).join();
+              const kbMatchesIdents = buckets[bucketKey][key].map((kbM) => kbM.ident).join();
+              const coalescedKbMatchesIdents = value.map((kbM) => kbM.ident).join();
               if (kbMatchesIdents !== coalescedKbMatchesIdents) {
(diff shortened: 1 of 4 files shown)


app/views/ReportView/components/KbMatches/coalesce.ts
```diff
@@ -114,7 +100,13 @@ const coalesceEntries = <T extends KbMatchedStatementType[]>(entries: T): Coales
       } else {
         Object.entries(entry).forEach(([key, value]) => {
           if (Array.isArray(buckets[bucketKey][key])) {
-            if (!buckets[bucketKey][key].includes(value)) {
+            if (key === 'kbMatches' && Array.isArray(value)) {
+              const kbMatchesIdents = buckets[bucketKey][key].map((kbM) => kbM.kbVariantId).join();
+              const coalescedKbMatchesIdents = value.map((kbM) => kbM.kbVariantId).join();
```
elewis2, line 105: I don't think we can coalesce on kbVariantId - two kbMatch records can have the same kbVariantIds but different observed variants. I think you could do all of (variant_type, variant_id, kb_variant_id) or kbmatch.ident

elewis2 (changes requested):

app/views/ReportView/components/KbMatches/coalesce.ts
bnguyen-bcgsc: Let me do a combination of kbVariantId and variantId. Using the ident might confuse the user when the 2 kbmatches have identical observed variants

bnguyen-bcgsc: @elewis2 actually let's keep using ident since the risk of having a different observed variant with the same kbVariantId and variant type is too high. Cannot rely on variant ident either since a lot of kb variant types don't have variant

elewis2 (approved):

kttkjl (approved):

PR #635, opened 2025-04-18
[DEVSU-2604] variant text bugfix on permissions

kttkjl:
Authenticated route uses a better system in determining re-route to mainpage -- uses `[something]Access` from useResources instead


kttkjl:
Should probably hard type the `requiredAccess` prop, `string` isn't enough

bnguyen-bcgsc (approved):

elewis2:
couple of bugs:
- when a user's info is successfully updated, I get two popups - one for the update and one saying I created the user successfully (but the user already exists)
- when a user has variant-text access, all the expected left-hand links are available (above the line). when a user doesn't have variant-text access (and instead has no special access roles), the only two links visible are to graphkb and to the terms and conditions. search and my-reports should still be accessible here

elewis2 (changes requested):
(changes in comment)

kttkjl:
> couple of bugs:
> 
> * when a user's info is successfully updated, I get two popups - one for the update and one saying I created the user successfully (but the user already exists)

DEVSU-2633 should take care of this

> * when a user has variant-text access, all the expected left-hand links are available (above the line). when a user doesn't have variant-text access (and instead has no special access roles), the only two links visible are to graphkb and to the terms and conditions. search and my-reports should still be accessible here

Can't reproduce, tried on `jacli-test` and removed all project/group permissions except `variant-text-edit`, has following:
- reports
- my-reports
- search
- graphkb link
- terms
- projects
- variant-text

Though no entries of variant-text returned from the API

elewis2:
As just discussed irl - I also can't reproduce this bug, will make a ticket if it happens again.

elewis2 (approved):

PR #636, opened 2025-04-23
[DEVSU-2635] Rapid/VariantEditDialog logic fix

kttkjl:
Logic for checking if `kbMatchedStatements` on a drug deletion were incorrect, logic loop kept looping itself causing all the ids to be up for deletion
  - fixed ids to delete logic on drug updates for RapidVariantEditDialog
  - rename Rapid/VariantEditDialog to RapidVariantEditDialog to avoid ambiguity with common component 
  - fixed some tabbing problems in code


elewis2 (approved):
(approved though ultimately the idea is not to delete these at all)

PR #641, opened 2025-05-06
[DEVSU-2620] [DEVSU-2634] rapid summary updates

kttkjl:
[DEVSU-2620]
Don't delete entries when removing from summary

[DEVSU-2634] 
Removed duplicate therapies on display

### DIFF
--- a/app/common.d.ts
+++ b/app/common.d.ts
@@ -179,6 +179,7 @@ type KbMatchedStatementType<T extends KbMatchType = KbMatchType> = {
     inferred: boolean;
     recruitment_status: string;
     kbmatchTag: string | null;
+    rapidReportTableTag: 'therapeutic' | 'cancerRelevance' | 'unknownSig' | 'noTable';
   } | null;
   kbMatches: T[];
   kbStatementId: string;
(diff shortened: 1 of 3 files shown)


elewis2:
Noting that this change means the mutation is not removed from the rapid report table when all treatments are hidden, leaving a variant row with 'sensitivity to' and nothing else displayed. Since there are more changes coming to the rapid reports summary kbmatches before next release I'm approving this pr anyway.

elewis2 (approved):

elewis2 (changes requested):
Found a bug - can't add a comment for high mutation burden variant type. maybe disable this field for that variant type (I believe it's getting processed as a signature variant) until we have a fix for that? alternatively maybe the comment can be stored in the observed-variants-annotations json field

kttkjl:
> Found a bug - can't add a comment for high mutation burden variant type. maybe disable this field for that variant type (I believe it's getting processed as a signature variant) until we have a fix for that? alternatively maybe the comment can be stored in the observed-variants-annotations json field

As discussed irl, let's make the front-end changes to check for the new type and then wait for `observed-variants-annotations` json field

elewis2 (approved):

bnguyen-bcgsc (approved):

PR #648, opened 2025-05-14
[DEVSU-2629] bulk add users by project

kttkjl:


bnguyen-bcgsc (changes requested):
I got a runtime error when trying to add project. Adding screenshot in Jira ticket

bnguyen-bcgsc (approved):

PR #665, opened 2025-07-28
DEVSU-2689 Rework Report History

bnguyen-bcgsc:
- DEVSU-2689
- Rename Analysis settings section to Report History
- Remove unused functionality of starting analysis and display report history
- Add additional report history milestones and sort them asc
- Update api querying to react-query

_Require DEVSU-2689 api change to work_


elewis2 (approved):

app/views/ReportView/components/Settings/components/ReportHistory/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const { report } = useContext(ReportContext);
+  let { reportEditAccess: canEdit } = useResource();
+  const [orderedHistory, setOrderedHistory] = useState([]);
+  if (report.state === 'completed') {
+    canEdit = false;
+  }
+
+  const { data: signatureHistoryData } = useQuery(
+    `/reports/${report.ident}/signatures/history`,
+    async ({queryKey: [route] }) => await api.get(route).request(),
```
kttkjl, line 24: linting flags this too

```

  const { data: signatureHistoryData } = useQuery(
    `/reports/${report.ident}/signatures/history`,
    async ({ queryKey: [route] }) => api.get(route).request(),
    {
      staleTime: Infinity,
      select: (response) => response,
    },
  );
```

is more inline with our lint rules

app/views/ReportView/components/Settings/components/ReportHistory/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+      staleTime: Infinity,
+      select: (response) => {
+        return response
+      },
+    },
+  );
+
+  const { data: stateHistoryData } = useQuery(
+    `/reports/${report.ident}/state-history`,
+    async ({queryKey: [route] }) => await api.get(route).request(),
```
kttkjl, line 35: same thing with the `return-await` rule we have here too

app/views/ReportView/components/Settings/components/ReportHistory/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+    },
+  );
+
+  useEffect(() => {
+    if (signatureHistoryData && stateHistoryData && report.users) {
+      const historyArr = signatureHistoryData.concat(stateHistoryData).concat(report.users);
+      historyArr.sort(function(a,b){
+        return new Date(a.updatedAt).valueOf() - new Date(b.updatedAt).valueOf();
+      });
+      const orderedHistoryArr = handleOrderReportHistory(historyArr);
```
kttkjl, line 50: `handleOrderReportHistory` is being used before being defined here

app/views/ReportView/components/Settings/components/ReportHistory/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+      setOrderedHistory(orderedHistoryArr);
+    }
+  }, [signatureHistoryData, stateHistoryData, report.users]);
+
+  const handleOrderReportHistory = (orderedHistory) => {
```
kttkjl, line 55: since this is being used in a `useEffect`, it would be worth it to wrap this into a `useCallback`, even with an empty dependency array, because React will know to not re-run the `useEffect` whenever it re-renders,

app/views/ReportView/components/Settings/components/ReportHistory/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+    const historyArray = [];
+    let authorSigned = null;
+    let creatorSigned = null;
+    let reviewerSigned = null;
+
+    for (const element of orderedHistory) {
+      if (element.hasOwnProperty('authorSignature')) {
```
kttkjl, line 62: More linting really
`Object.prototype.hasOwnProperty.call(element, 'authorSignature')` is a workaround for the [no-prototype-builtins](https://eslint.org/docs/latest/rules/no-prototype-builtins) rule, but we COULD update the compiler option in `tsconfig.json` to `"lib": ["ES2022", "DOM"]` so then you can use `Object.hasOwn(element, 'authorSignature')` instead

app/views/ReportView/components/Settings/components/ReportHistory/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+            <br />
+          {report?.analysisStartedAt ? (
+            <>
+              Analysis started on:
+              {` ${formatDate(report?.analysisStartedAt, true)}`}
+              <br />
+            </>
+          ) : null}
+          {orderedHistory ? orderedHistory.map((result, index) => (
+            <div key={index}>
```
kttkjl, line 117: indexes are always bad idea for `key` in React as well, it gets into the way of DOM updates, since `id` is supposed to be unique

kttkjl (changes requested):
Linting problems, unused variables too

PR #667, opened 2025-08-02
Bugfix/devsu 2692 rapidsummary kbmatches variants fix

kttkjl:


kttkjl:
Hmmm, there's a bug atm, fixing

elewis2:
Can statements where there are no therapies to be displayed, be hidden? I'm seeing eg 'sensitivity to' and then nothing in the therapies column. Similarly if the statement's context is not a therapy the context should not be displayed - so eg 'eligibility' should not be displayed (this requirement could change in future but if so we'd want to display the trial name with it; at this point this feature has not yet been requested).

kttkjl:
@bnguyen-bcgsc do you have a report ID so I can see about replicating it, thanks

kttkjl:
> 2. From KbMatches, ABCB1 row (sensitivity) cannot be added to the rapid summary top row but other variants can

On surface, this looks like an API error, as the POST request is failed, the PUT request for the same variant is successful, but somehow the annotation is not being shown

elewis2 (changes requested):
As discussed on rc, there is a bug to work out re observed-variant-annotation post vs put; as noted in comment also don't display statements that don't relate to a therapy

bnguyen-bcgsc:
> Enabled the comment API call on the client, but it seems like sigv variants do not allow comments to be edited ` "message": "Error while validating signatureVariants update request Error: should NOT have additional properties [additionalProperty: comments]"`

I just checked the API, looks like the sigV model does not have a comments field like the other variant types do. API probably needs to be updated

kttkjl:
TODO: when a kbMatchesStatement is aggregated on kbMatches page, make it clear which variant's matched statement is being added instead of aggregating both

sshugsc (dismissed):

bnguyen-bcgsc (dismissed):

elewis2 (dismissed):

PR #678, opened 2025-09-18
Feat/devsu 2692 pt 2 filter all but selected stmt

elewis2:
Uses new endpoint to handle updating the rapid summary table for a single variant and a list of associated statements. 

Also adds a default value for iprEvidenceLevel as this caused unexpected behaviour during testing.

Works with this pori_ipr_api branch:

bugfix/DEVSU-2724-only-add-single-statement-to-rapid-summary

### DIFF
--- a/app/views/GermlineView/components/Board/components/ApiPaginatedTable/components/CheckboxCell/index.tsx
+++ b/app/views/GermlineView/components/Board/components/ApiPaginatedTable/components/CheckboxCell/index.tsx
@@ -3,6 +3,7 @@ import {
   Checkbox,
 } from '@mui/material';
 import { useSnackbar } from 'notistack';
+
 import { ICellRendererParams } from '@ag-grid-community/core';
 
 import api from '@/services/api';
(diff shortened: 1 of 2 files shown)


app/views/ReportView/components/KbMatches/index.tsx
```diff
@@ -60,7 +60,7 @@ const KB_MATCHES_TITLE_MAP = {
 
 const RAPID_TABLE_TITLE_MAP = {
   // Should be therapeuticAssociation, but the tag in backend is looking for 'therapeutic'
-  therapeutic: 'Variants with Clinical Evidence for Treatment in This Tumour Type',
+  therapeuticAssociation: 'Variants with Clinical Evidence for Treatment in This Tumour Type',
```
kttkjl, line 63: the reason why it's `therapeutic` instead of `therapeuticAssociation` was due to 
```
    "kbMatches": [
        {
            "category": "therapeutic",
      ...
```

kttkjl (approved):

PR #765, opened 2026-04-13
DEVSU-2917 Peds CD8+ TCell Fields Fixes

bnguyen-bcgsc:
- DEVSU-2917
- Fix issue where rapid and genomic summary would not display peds cd8+ tcell fields
- Fix issue where peds cd8+ tcell comment field is disabled with score field being 0.0 instead of null


elewis2 (dismissed):

app/components/TumourSummaryEdit/index.tsx
```diff
@@ -209,6 +209,9 @@ const TumourSummaryEdit = ({
       ...cd8t,
       [name]: parseFloat(value),
     }));
+    if (value === null || value === '') {
```
kttkjl, line 208: shouldn't this guard go first and then parseFloat? like so

```
if (value === null || value === '') {
  setNewTCellCd8Data((cd8t) => ({ ...cd8t, [name]: null, pedsScoreComment: null }));
} else {
  setNewTCellCd8Data((cd8t) => ({ ...cd8t, [name]: parseFloat(value) }));
}
```

should `value` be `null` we'll end up with `NaN`, of course, the current if statement will catch it, but why even set in the first place if it's an invalid value

app/components/TumourSummaryEdit/index.tsx
```diff
@@ -783,8 +786,8 @@ const TumourSummaryEdit = ({
         label="Pediatric CD8+ T Cell Comment"
         value={newTCellCd8Data?.pedsScoreComment ?? ''}
         name="pedsScoreComment"
-        disabled={!newTCellCd8Data?.pedsScore && !newTCellCd8Data?.pedsScoreComment}
-        required={!!newTCellCd8Data?.pedsScore}
+        disabled={newTCellCd8Data?.pedsScore === null && !newTCellCd8Data?.pedsScoreComment}
```
kttkjl: this change makes it that whenever `pedsScore` is `''` (which if you enter a value and then clear it in the UI) the comment is still editable

the only time in which this is disabled is when the value comes back from the API is `null`

app/components/TumourSummaryEdit/index.tsx
bnguyen-bcgsc: awesome I will update this

app/components/TumourSummaryEdit/index.tsx
bnguyen-bcgsc: will update!

app/views/ReportView/components/RapidSummary/index.tsx
```diff
@@ -393,8 +393,8 @@ const RapidSummary = ({
     }
 
     let tCell: null | string;
-    if (tCellCd8 && typeof tCellCd8.score === 'number') {
-      if (tCellCd8.pedsScore) {
+    if (tCellCd8 && (typeof tCellCd8.score === 'number' || typeof tCellCd8.pedsScore === 'number')) {
```
kttkjl: would it be cleaner to short circuit the logic first here

```
const hasPedsScore = typeof tCellCd8?.pedsScore === 'number';
```

```
// outer guard (unchanged — already checks both)
if (tCellCd8 && (typeof tCellCd8.score === 'number' || hasPedsScore)) {
  if (hasPedsScore) {
    tCell = `${tCellCd8.pedsScore} ...`;
  } else {
    tCell = `${tCellCd8.score} ...`;
  }
}

// label
term: hasPedsScore ? 'Pediatric CD8+ T Cell Score' : 'CD8+ T Cell Score'

// comment
value: tCellCd8?.pedsScoreComment && hasPedsScore ? tCellCd8.pedsScoreComment : null
```

app/views/ReportView/components/RapidSummary/index.tsx
```diff
@@ -393,14 +393,13 @@ const RapidSummary = ({
     }
 
     let tCell: null | string;
```
kttkjl: missing init of `tCell = null` ?

app/views/ReportView/components/GenomicSummary/index.tsx
```diff
@@ -226,8 +226,8 @@ const GenomicSummary = ({
       }
 
       let tCell: null | string;
-      if (tCellCd8 && typeof tCellCd8.score === 'number') {
-        if (tCellCd8.pedsScore) {
+      if (tCellCd8 && (typeof tCellCd8.score === 'number' || typeof tCellCd8.pedsScore === 'number')) {
```
kttkjl: feels like this coding style should be the same as RapidSummary?, quite like the `hasPedsScore`

kttkjl (approved):

PR #775, opened 2026-05-01
[DEVSU-2886] TO tables notes multiline

kttkjl:
To be added in after the release 

Seems like storybook isn't a dev requirement, installed that too


elewis2 (changes requested):
see comment

elewis2 (approved):

PR #779, opened 2026-05-08
Bugfix/devsu 2943 expression page broken

bnguyen-bcgsc:
- DEVSU-2943
- Update deprecated getAllColumns function from new ag grid version
- Add checks for potential empty columns array from colApi getColumns
- Remove duplicated column auto sizing useEffect in DataTable since onFirstDataRendered is already taking care of that, preventing issue where colApi is attempting to resize columns before useEffect finish rendering columns
- Update test suites to not throw when visibleColumns are undefined in case colApi fires util functions before data gets rendered

### DIFF
--- a/app/components/DataTable/index.tsx
+++ b/app/components/DataTable/index.tsx
@@ -296,17 +296,6 @@ const DataTable = forwardRef<DataTableImperativeHandle, DataTableProps>(({
     }
   }, [filterText, gridApi]);
 
-  // Triggers when syncVisibleColumns is called
-  useEffect(() => {
-    if (colApi && visibleColumns.length) {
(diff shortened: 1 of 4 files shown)


app/components/DataTable/index.tsx
```diff
@@ -296,17 +296,6 @@ const DataTable = forwardRef<DataTableImperativeHandle, DataTableProps>(({
     }
   }, [filterText, gridApi]);
 
-  // Triggers when syncVisibleColumns is called
```
areisle, line 299: why is this part being removed?

bnguyen-bcgsc: you're right I added it back in and used a different flag to check for data being rendered first

elewis2 (approved):

PR #788, opened 2026-05-14
[DEVSU-2951, DEVSU-2952] TO table updates

kttkjl:
Moved `canReorder` out to make `DataTable` more modular, because reordering might not be limited to just rank.

Updated tests 

Merged with DEVSU-2952


app/components/DataTable/index.tsx
```diff
@@ -142,12 +142,14 @@ export type DataTableImperativeHandle = {
...
   showLoading: () => void;
+  getGridApi: () => GridApi | null;
+  getColumnApi: () => ColumnApi | null;
 };
 
 type DataTableCustomProps = {
   /* Text shown next to the add row button */
   addText?: string;
-  additionalTableMenuItems?: (gridApi: GridApi) => JSX.Element | JSX.Element[];
+  additionalTableMenuItems?: (gridApi: GridApi, closeMenu: () => void) => JSX.Element | JSX.Element[];
```
elewis2, line 152: is the instance of additionalTableMenuItems in KbMatches affected?

bnguyen-bcgsc: oh yea it is affected on line 695 too (of the kbmatches view)

kttkjl: Yes, any instances that uses `DataTable` that has `additionalTableMenuItems` as a prop should be affected

bnguyen-bcgsc (approved):

PR #797, opened 2026-06-22
DEVSU-2980 Add Small Mutations to Summary (Genomic Alterations Identified)

bnguyen-bcgsc:
- DEVSU-2980
- Fix small mutations variantType when add to summary using VariantEditDialog

### DIFF
--- a/app/components/VariantEditDialog/index.tsx
+++ b/app/components/VariantEditDialog/index.tsx
@@ -47,7 +47,7 @@ const VariantEditDialog = ({
     if (editData) {
       let newVariant;
       switch (variantType) {
-        case 'snv':
+        case 'mut':
           newVariant = `${editData?.gene.name}:${editData?.proteinChange}`;
           setVariant(newVariant);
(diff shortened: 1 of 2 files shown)


app/components/VariantEditDialog/index.tsx
```diff
@@ -47,7 +47,7 @@ const VariantEditDialog = ({
     if (editData) {
       let newVariant;
       switch (variantType) {
-        case 'snv':
+        case 'mut':
```
kttkjl, line 50: this looks like a replace more than an add ..?

kttkjl: oh that's a typo isn't it due to `allowedValues`  from the API

bnguyen-bcgsc: yea it was a typo, the allowedValues has mut instead of snv for small mutations

kttkjl (approved):

PR #806, opened 2026-07-16
[DEVSU-2949] geneviewer display fixes

kttkjl:
Add pagination to GeneViewer tables
Columns now resizable
Columns now fit to data on first render


kttkjl:
> Right now, the horizontal scroll bar and pagination are only visible when the user scrolls all the way down to the end of the dialog. Is there a way we can make the table height fixed to the height of the dialog, so the vertical scrolling is table-level instead of dialog-level? This way the user can scroll horizontally while still seeing the top tabs and headers

yeah I thought about this too, should we check with clininfo on what they want?

bnguyen-bcgsc (approved):

kttkjl:
wrong ticket number, whoops

